PR #24, opened 2017-02-11
Allow special characters like Ampersand in WiFi password

DottoreTozzi:
Fixed form method in config portal in order to allow special characters in WiFi password (especially & was causing problems).


DottoreTozzi:
It works here. Enctype is only available for POST method...

universam1:
Thanks for the research!

I would love to keep the get method because it is very handy for debugging plus you can store your various configurations just by a link that's super handy for set ups. 
Looks like using HTML encode of special character does solve it also, have tot research how to implement.

universam1:
According to http://www.w3schools.com/Tags/ref_httpmethods.asp
the application/x-www-form-urlencoded is supported also for a get method, this should fix our issue. 
Could you give it a try please?

DottoreTozzi:
I understand why you wish to keep GET, but unfortunately it won't work; I've tried.
application/x-www-form-urlencoded is the default, and should be the correct enctype, but for some reason this doesn't even work with POST, even though it theoretically by all means should, at least according to my understanding.

The WPA2-WPK is being truncated after an & is encountered, however.
For example, "yZf65&pxR48%" becomes "yZf65", even though it should be passed as "yZf65%26pxR48%25".

It must have something to do with the way the ESP8266WebServer handles arguments.
My guess is that, apparently, it decodes the whole argument string first and then parses it, instead of correctly doing this the other way round.
I'll do some further research on this but that's going to be very time consuming, I'm afraid.
For now, I don't see any other way to fix this issue. But I'll look into it.

OTOH, the POST method is the recommended way to handle sensitive stuff like passwords, anyway...

DottoreTozzi:
OK, this does the trick:

```cpp
--- a/libraries/ESP8266WebServer/src/Parsing.cpp
+++ b/libraries/ESP8266WebServer/src/Parsing.cpp
@@ -91,7 +91,7 @@ bool ESP8266WebServer::_parseRequest(WiFiClient& client) {
   String searchStr = "";
   int hasSearch = url.indexOf('?');
   if (hasSearch != -1){
-    searchStr = urlDecode(url.substring(hasSearch + 1));
+    searchStr = url.substring(hasSearch + 1);
     url = url.substring(0, hasSearch);
   }
   _currentUri = url;
@@ -318,7 +318,7 @@ void ESP8266WebServer::_parseArguments(String data) {
     }
     RequestArgument& arg = _currentArgs[iarg];
     arg.key = data.substring(pos, equal_sign_index);
-       arg.value = data.substring(equal_sign_index + 1, next_arg_index);
+    arg.value = urlDecode(data.substring(equal_sign_index + 1, next_arg_index));
 #ifdef DEBUG_ESP_HTTP_SERVER
     DEBUG_OUTPUT.print("arg ");
     DEBUG_OUTPUT.print(iarg);
```

universam1:
https://github.com/esp8266/Arduino/pull/2956

PR #26, opened 2017-02-11
config-mode: ESP "reset-reason" workaround

JensW2000:
The ESP reset info is sill buggy. see http://www.esp8266.com/viewtopic.php?f=32&t=8411
The reset reason is "5" (woken from deep-sleep) in most cases (also after a power-cycle)
I added a single reset detection as workaround to enter the config-mode easier


universam1:
Did you test whether it conflicts with normal mode reboots?
 Had this checks in firmware v1 but dropped since it was not a reliable test.

JensW2000:
I tested it on 5 iSpindles.

With this change the behavior after a power-cycle is:
- sometimes the ESP API reports "6" as boot reason and the iSpindle turns into config-mode
- it's not reliable, bebause the ESP API returns the boot reason "5" (deep-sleep awake)

The behavior after a single reset is reproducible in any cases:
- the iSpindle is turning into Config-Mode after a single reset

The pull request contains both methods. A single reset will enter the config-mode in any cases. A power-cycle will enter the config-mode sometimes (as long as the ESP API bug hasn't been fixed).
Neither a normal deep-sleep awake nor a soft-reset (after saving the config) will cause a config-mode boot.

Looks fine for me.

universam1:
Nice, thanks for working this out! Looks good to me, going to merge

PR #28, opened 2017-02-13
Update compile.md

DottoreTozzi:
Added warning about Board Manager.
I ran into problems because of this and as it turns out, others are, too...
So I thought perhaps we should state this more clearly.


universam1:
Valuable suggestion - thank you!

PR #29, opened 2017-02-15
Fix: Random crash on entering config portal from config mode due to PROGMEM inconsistencies

DottoreTozzi:
There was a weird condition when compiled code would crash the ESP9266, depending on what host OS and/or Arduino IDE version it was compiled on.    
After lots of debugging, it turns out this has to do with PROGMEM.    
The WiFiManagerParameter api_list has no properties set besides HTTP_API_LIST, which seems to (sometimes) cause problems when directly referencing it, being then (sometimes) read incorrectly from flash memory and causing a type 3 exception.    
This **very** hard-to-find fix should restore compatibility across platforms and IDE versions.    
Might want to keep this in mind, should similar problems arise with other "v0g0n" PROGMEM char[]s in the future, fingers crossed.

### DIFF
--- a/docs/compile.md
+++ b/docs/compile.md
@@ -13,3 +13,9 @@
 
 - https://github.com/esp8266/Arduino
 - https://github.com/bblanchon/ArduinoJson
+
+## Note
+
+Do **not** use the Board Manager to install the ESP8266 core.    
+Clone from github into ~/Documents/hardware instead, as described in the documentation.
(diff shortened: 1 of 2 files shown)


universam1:
Nice job @DottoreTozzi with reinterpreting the PROGMEM Flash. Looks like the compiler acts differently between OSes. I remember having a weird issue with Flashstringhelper before, have to find that place again and try that explicit definition.
Suggest blaming it upward to the original library!

PR #33, opened 2017-02-20
Added Installation instructions for Raspbian

DottoreTozzi:


universam1:
Nice Howto - thanks for the work

PR #46, opened 2017-03-16
genericTCP for Local Server: Added Firmware 5.x Support, Fixed Plato Chart, Updated Docs

DottoreTozzi:
Backwards compatible Update to genericTCP and charts
Firmware 5.x compatibility
Docs updated


universam1:
S W E A T

Issue #48, opened 2017-03-18
`getTemparature()` causes system crash and prevents saving of configuration


ahanin:
While saving configuration the program would crash with following stacktrace:

    Exception (9):
    epc1=0x40105a6d epc2=0x00000000 epc3=0x00000000 excvaddr=0xffffffff depc=0x00000000
    
    ctx: sys 
    sp: 3ffffdb0 end: 3fffffb0 offset: 01a0
    
    >>>stack>>>
    3fffff50:  00000056 00000001 401051ee 3ffee030  
    3fffff60:  4021cdbd 3ffee8e8 00000001 00000000  
    3fffff70:  402183eb 00000004 00000000 00000000  
    3fffff80:  3fff1cdc 40217de2 3fffdab0 00000000  
    3fffff90:  3fffdcc0 3ffe9f60 00000000 4022fd4a  
    3fffffa0:  3ffe9f60 40000f49 3fffdab0 40000f49  
    <<<stack<<<

This translates to:

    Exception 9: LoadStoreAlignmentCause: Load or store to an unaligned address
    
    0x40105a6d: ets_timer_disarm at ?? line ?
    0x401051ee: lmacTxFrame at ?? line ?
    0x4021cdbd: ieee80211_getmgtframe at ?? line ?
    0x402183eb: ppCheckTxIdle at ?? line ?
    0x40217de2: ppPeocessRxPktHdr at ?? line ?
    0x4022fd4a: wifi_set_status_led_output_level at ?? line ?

The error causes the board to reset and configuration is not saved. I observed it only in combination with WiFi functionality. 

I use Amica ESP8266 NodeMCU 1.0 board connected via USB without a temperature probe or any other devices connected. The stacktrace varies, but it always happens in `ets_timer_disarm` routine. I looked at what is running in timer, and it was the `flash` function calling `getTemperature()` amongst the others. And there there is a `delay(200)`. I am not sure about the reason for this delay. Perhaps a minimal waiting time before DS boots up. But if I remove it completely or reduce to 100ms, the error is not being reproduced and configuration is changed.  

I understand that I don't use D1 mini, but they are the same chip. Also, I don't have temperature probe connected which probably causes the problem. But apparently there is an unfortunate combination of events that causes that and prevents from tinkering around with the code.

Proposed pull request: #49 

[...]

PR #49, opened 2017-03-18
#48 reducing delay after turning OneWire off

ahanin:


universam1:
It is not actually a bug if peripheral is not installed

PR #50, opened 2017-03-22
Add files via upload

Henielma:
Hereby a Dutch description for assembling a iSpindel using a breadboard.


universam1:
do you want to reference it maybe from the readme_nl.md?

Henielma:
A good idea i think. If you can do that its fine for me.

PR #64, opened 2017-05-26
Use ESP8266 pin numbers

chriszero:
allows to use different ESP8266 boards without changing code. e.g.
Adafruit HUZZAH ESP8266

### DIFF
--- a/iSpindel/Globals.h
+++ b/iSpindel/Globals.h
@@ -52,7 +52,7 @@ extern Ticker flasher;
 #define PORTALTIMEOUT 300
 
 #define ADCDIVISOR 191.8
-#define ONE_WIRE_BUS D6  // DS18B20 on ESP pin12
+#define ONE_WIRE_BUS 12 // on Wemos D1 mini pin D6
 #define RESOLUTION 12 // 12bit resolution == 750ms update rate
 #define OWinterval (800 / (1 << (12 - RESOLUTION))) 
 #define CFGFILE "/config.json"
--- a/iSpindel/iSpindel.ino
+++ b/iSpindel/iSpindel.ino
@@ -555,7 +555,7 @@ void initDS18B20()
 void initAccel()
 {
   // join I2C bus (I2Cdev library doesn't do this automatically)
-  Wire.begin(D3, D4);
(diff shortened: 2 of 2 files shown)


universam1:
This project is very focused on the Wemos D1 mini. Changing the pin numbers to ESP layout native is technically correct but confusing for the ones who try to understand the schematic. 
Hence at least should be a comment about their adoption.

chriszero:
Wemos is nothing else than a breakout for the ESP8266...
I have added some comments to the previous commit.
I've build a iSpindel with an Adafruit HUZZAH ESP8266: http://hobbybrauer.de/forum/viewtopic.php?p=215520#p215520
Works great, if wanted I can write a wiki page.

universam1:
Sure, though it will be confusing.
Now, as we start to support alien boards let's make it right then as the Huzzah is by far not the only interesting ESP8266 breakout board. 
A proper implementation would be to introduce a #ifdef preprocess block on board types in config.h
Hence the layout becomes free as well.

chriszero:
I think there is no need for board specific preprocessor blocks, as the pins of the ESP never change. The most boards have the pin numbers 1:1

jsnwtsn:
@chriszero - I think a wiki for using the Adafruit Huzzah would be great. I have a few Huzzahs and would like to use them for this project.

rlcout:
I tried use NodeMcu esp8266 12-E but hey don´t save the wifi network and get deep sleep.  Anybody know whats going on?

universam1:
No activity any more so closing

jsnwtsn:
@universam1 I was hoping the Huzzah would have gotten more support since it has a built in LiPo charger. Paired with the ultralow power ADXL345, it might have made for an good hardware alternative.

@chriszero Do you have notes on your Huzzah/iSpindel build that you could share?

universam1:
I’m fine with adding support for other boards. A proper implementation would look like adding the board to platformio and guarding that case in the pin definition. Just switching the pin names is dangerous as a build for Wemos might work or break the Huzzahs flash layout.

chriszero:
I will look into it, and clean the code.

universam1 <notifications@github.com> schrieb am Do., 14. Dez. 2017, 11:13:

Issue #116, opened 2017-11-17
Send interval


cwilling:
Currently there is no way for a server to know when the next data transmission from an iSpindel will occur. Some services (like Ubidots which just display graphs of whatever data have been received) don't care - they just keep waiting.

Other services may actually use the received data for something important e.g. making beer. In this case, the server might be switching an external device like a fridge on or off according to some algorithm which bases its decisions on the data received from an iSpindel. A question arises of what the algorithm should do when (for whatever reason) transmissions from the iSpindel stop.

I don't want to discuss now what the algorithm should do in such a case. What I would like to discuss is how the server can detect that contact with the iSpindel has stopped. In my own software, I have a configuration setting so that the user can set a time limit such that if the server does not receive a data transmission within the configured time limit, it assumes the iSpindel has died. I generally advise this setting to be a a small amount (say 5%) over two times the sleep interval of the iSpindel itself e.g.  sleep interval 300s => time limit 630s. The rationale is that we may forgive the loss of 1 data transmission but if we lose two consecutive transmissions, we should start to worry that something is wrong.

The manual configuration method works well enough but a problem arises when a user has configured the iSpindel with a quite short sleep interval (maybe for calibration or testing etc.) and later reconfigures the iSpindel for a longer sleep interval, perhaps to start some real work, but forgets to set a new time limit in the server. 

Due to a manual oversight like this, the software now doesn't wait long enough for data transmissions and wrongly assumes it has to deal with a dead iSpindel. It would be better not to depend on the user always doing the right thing (remembering to update the software time limit to match the iSpindel). It would be far preferable if the timeout could be set programmatically. This would require that the software can know what the iSpindel sleep interval has been configured to be. The software could know this if the iSpindel would include the sleep interval in its regular data transmissions.

I have modified the iSpindel firmware locally to include the sleep interval and it works quite well. I will submit a PR with the changes shortly. In the meantime, are there any comments about the situation? Any other solutions for the same problem?

universam1:
In general, this is a pretty specific issue, I am not aware of other software that is considering an update health. 
Nevertheless, I think your reasoning is good and I like the simplicity to solve this challenge, we should merge it.

Now there are a few doubts about some protocols that are proprietary and will likely break, I suggest we leave them out for now. If the stakeholder decide to support we can add them individually anytime.

PR #117, opened 2017-11-17
Include sleep interval in data transmissions

cwilling:
This change enables receiving software to know the configured sleep time, addressing issue #116 

Signed-off-by: Christoph Willing <chris.willing@linux.com>

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -421,7 +421,6 @@ bool uploadData(uint8_t service)
     sender.add("temperature", Temperatur);
     sender.add("battery", Volt);
     sender.add("gravity", Gravity);
-    sender.add("interval", my_sleeptime);
     SerialOut(F("\ncalling Ubidots"));
     return sender.sendUbidots(my_token, my_name);
   }
@@ -474,7 +473,6 @@ bool uploadData(uint8_t service)
     sender.add("battery", Volt);
     sender.add("gravity", Gravity);
     sender.add("ID", ESP.getChipId());
-    sender.add("interval", my_sleeptime);
     SerialOut(F("\ncalling FHEM"));
     return sender.sendFHEM(my_server, my_port, my_name);
   }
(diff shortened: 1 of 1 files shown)


pio/src/iSpindel.cpp
```diff
@@ -483,6 +486,7 @@ bool uploadData(uint8_t service)
     sender.add("D", Tilt);
     sender.add("U", Volt);
     sender.add("G", Gravity);
+    sender.add("I", my_sleeptime);
```
universam1: this one is not a flexible protocol so it will break. Hence unnecessary here.

pio/src/iSpindel.cpp
```diff
@@ -472,6 +474,7 @@ bool uploadData(uint8_t service)
     sender.add("battery", Volt);
     sender.add("gravity", Gravity);
     sender.add("ID", ESP.getChipId());
+    sender.add("interval", my_sleeptime);
```
universam1: I'm not sure about FHEM if it is additional data tolerant, I would rather not risk to break.

pio/src/iSpindel.cpp
```diff
@@ -421,6 +421,7 @@ bool uploadData(uint8_t service)
     sender.add("temperature", Temperatur);
     sender.add("battery", Volt);
     sender.add("gravity", Gravity);
+    sender.add("interval", my_sleeptime);
```
universam1: as @cwilling you mentioned, ubidots is ignorant to this challenge. I dont have a strong feeling here, but consider this additional data is billed.

cwilling:
I've made the changes you suggested and pushed to my "interval" branch on cwilling/iSpindel. I believe that should automatically update this pull request.

pio/src/iSpindel.cpp
cwilling: This comment about ubidots seems to be attached to code line 441 which is actually part of API_GENERIC. In my update to the PR, I removed extra interval field from API_UBIDOTS.

universam1: correct, thanks

universam1:
yeah sounds good to me, thanks for the contribution!

PR #144, opened 2018-02-19
Increase tokensize to 40

### LINKED ISSUE #143: New tokensize ubidots
I've just created an Ubidots account and my token is in the following format:
```
Educational: A1E-abcdefghijklmnopqrstuvwxyzabcd
Business:    BBFF-abcdefghijklmnopqrstuvwzyzabcd
             123456789012345678901234567890123456
```
This token itself is the specified 35 length but we also need an "\0" to end the string.

Edit: Stupid me, applied for a Business Ubidots account which do carry longer prefixes

Stapelpanda:
Fixes #143


universam1:
Why?

universam1:
Okay I see the point. Fine with me.  We though need to bump the version to make it accessible.

PR #147, opened 2018-02-25
Add support for logging data to a influxdb database server

iceaway:
Initital support for logging data to an influxdb database server. Future improvements could be configurable "table" in the influx db, as well as the possibilty to check if the database exists already, and in that case create it.


tedelm:
+1

PR #155, opened 2018-03-04
Implementation of https interface / JSON Page "/sensors" / minor bug fixes / 

mestrode:
* Implementation of https interface
  * https interface
  * modify token size to hold an 40 char fingerprint + delimiter
* JSON page /sensors in config mode (get all sensor values at once in as JSON data)
* some minor changes due to code style
  * fix of http (merge of url string)
  * resort of some codelines to make a better overview (iSpindel.cpp, WiFiManagerTK, )
  * comments for some functions added
  * input formular: use of length vs. maxlength
  * fix: is you use #define should use /*comments*/ not // comments

### DIFF
--- a/pio/lib/Globals/Globals.h
+++ b/pio/lib/Globals/Globals.h
@@ -39,7 +39,7 @@ extern Ticker flasher;
 #define RESOLUTION 12 // 12bit resolution == 750ms update rate
 #define OWinterval (800 / (1 << (12 - RESOLUTION))) 
 #define CFGFILE "/config.json"
-#define TKIDSIZE 40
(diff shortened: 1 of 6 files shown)


universam1:
The intention of this PR is fine but it violates the generic pattern of contributions. 
You are not to submit several changes with a single request unless functionally wise unavoidable. 
The reason is simple that other developments are ongoing and will conflict. 
A general DevOps principle is commit small, atomic and often. 
Refactor this PR then we can work on merging. 

However I don’t favor a special function for SSL connection it should be possible to smarter handle this case.

mestrode:
rework

PR #159, opened 2018-03-07
Move strings to FLASH to save RAM

FernandoGarcia:
I haven't enough hardware to make a full test with this code but I did upload to a nodeMCU and I have accessed the portal without problems.  It should save about 1550 bytes of RAM.

Sorry if I'm doing something wrong but this is my first PR. :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ iSpindel/__vm/
 .vscode
 .vscode/c_cpp_properties.json
 .vscode/launch.json
+.clang_complete
+.gcc-flags.json
--- a/pio/lib/Globals/Globals.h
+++ b/pio/lib/Globals/Globals.h
@@ -38,14 +38,14 @@ extern Ticker flasher;
 #define ADCDIVISOR 191.8
 #define ONE_WIRE_BUS D6  // DS18B20 on ESP pin12
 #define RESOLUTION 12 // 12bit resolution == 750ms update rate
-#define OWinterval (800 / (1 << (12 - RESOLUTION))) 
-#define CFGFILE "/config.json"
+#define OWinterval (800 / (1 << (12 - RESOLUTION)))
+#define CFGFILE F("/config.json")
 #define TKIDSIZE 40
 #define MEDIANROUNDS 7
 #define ACCINTERVAL 200
 #define MEDIANAVRG 3
 
-#define CBP_ENDPOINT "/api/hydrometer/v1/data"
+#define CBP_ENDPOINT F("/api/hydrometer/v1/data")
 
 #define DTUbiDots 0
 #define DTThingspeak 1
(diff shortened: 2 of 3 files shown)


universam1:
In general this is fine with me, but there are a number of situations where PROGMEM strings caused exceptions in the run, that are by the way really ugly to debug.
If the latest SDK works now reliable with FLASH strings would be nice, but as I have a history with it I ask you to test it thoroughly firsthand.

FernandoGarcia:
Will take long time since parts are coming from China. If someone else can test it could be cool.

FernandoGarcia:
Each char in FLASH is 1 byte more of free RAM.

universam1:
Well, that would be a great world, right. Unfortunately, that is not the case.

Go out and try to compile:
`Serial.print("X");`
vs
`Serial.print(F("X"));`

Surprise, you don't save that byte, why? Because you have to know how this Flashstring works that there is some cost with it.. Actually, I needs to create a buffer of 4 byte array in runtime first, access the 4x Byte == 32bit Flash and load into buffer, than return this pointer to the function. 

As you see the PROGMEM access comes with a cost, at least in processor cycles, slow flash access wait states and also in RAM buffer. So from what point does it pay off? 
There are a number of discussions out there when it makes sense to use `F()`, and for sure not on 1 byte of data.

universam1:
Hey @FernandoGarcia for the sake of your efforts I created a MR for to use flashstrings where safe and also gain a few more space from dropping the generic handler
Check #164 if that is your intention.

universam1:
@FernandoGarcia I used that statement to refactor interesting places that pay off with >= 4 byte which is a very positive assumption.
```
(Serial[^F]+\()("[^"]{4,}")
$1F($2)
```

FernandoGarcia:
To check if save or no RAM just check the freeHeap on portal.

If you will remove **SerialOut** function as I told in #160 is better add a **#ifdef** to allow disable debug messages or disable the console as you prefers call.

universam1:
@FernandoGarcia #165 should explain you the concept

FernandoGarcia:
Sorry, I don't want bother you but to me SerialOut function was only renamed in #165.

To save RAM and FLASH when console and debug are disabled all prints should be inside of #ifdef DEBUG and #endif or #ifdef CONSOLE and #endif.

Like this:

```
#ifdef DEBUG  
      json.printTo(Serial);  
#endif  
```

universam1:
Never mind @FernandoGarcia 
Check this little compiler magic here https://github.com/universam1/iSpindel/pull/165/files#diff-21506c7b4bcb613f5179321d2dd5d798R36
This is effectively the same as your `#ifdef` statement, just more convenient!

FernandoGarcia:
Anyway it's obsolete now. Maybe you can implement the F("") to json keys yet.

PR #167, opened 2018-03-14
implementation https

mestrode:
finally...


universam1:
I unfortunately have to repeat it over again, a PR is to a be a one commit, one topic change. 
Your PR has 16 commits and covers almost the same amount of different ideas. Hence to many conceptional conflicts.

simduchaine:
Can you clean up this PR. I really need the https implementation..

universam1:
In meantime it would be the better way to go with BearSSL instead. Note the costs involved: https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/bearssl-client-secure-class.html#cpu-requirements
Hence going to close this PR

PR #170, opened 2018-03-27
Translated the calibration sheet and added column for SG values

knockimov:
Updated the calibration sheet to use SG values directly for an easier calibration

Changes:
+ New column for SG values
+ Plato field autofills with converted SG value
+ Moved formel field and highlighted it
+ Translated the visual texts

Uses:
Read tilt values from iSpindel and measure SG and enter those values in respective columns. No more need to convert SG values to Plato because the sheet automatically converts it for you.
When you have enough values entered, press the button "Formel update" and copy the new plato formel to your iSpindel config.


knockimov:
I have updated my post again with only the translation part.

Considering your suggestion about using platformIO instead of the code itself is the path to go. Thank you!

PR #187, opened 2018-08-05
Add temp units to output (Closes #186)

### LINKED ISSUE #186: Temperature units not sent
Following [these changes](https://github.com/universam1/iSpindel/commit/a082b62351e8a1a8b5c69fd7a81a7839f1f732fd#diff-f0fd4a34e1fb120d9c064b4cc883947e), temperatures are now sent to remote loggers in the units specified by the user (F, C, or K). What units the user has specified, however, is not. This prevents remote services from knowing what temperature scale the reading is on.

I would propose adding a new field to be sent - `temp_units` - that would contain the units of temperature in use.

thorrak:
With the recent addition of the ability to automatically convert the temperature being sent downstream to units other than Celsius, downstream receivers of iSpindel data could no longer determine the units of temperature they were being provided. As a result, for those consumers who had built conversion functionality, the assumption that the temperature being provided was in Celsius was no longer true. This results in users seeing [wildly inaccurate temperatures](https://www.homebrewtalk.com/forum/threads/native-esp8266-brewpi-firmware-wifi-brewpi-no-arduino-needed.586476/page-52#post-8362123) if units other than Celsius were selected on the iSpindel.

This pull request adds a new field that is sent along with the scaled temperature that contains a character label for the units of temperature in use (C, F, or K).

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -101,6 +101,19 @@ float scaleTemperature(float t)
     return t; // Invalid value for my_tempscale => default to celsius
 }
 
+String tempScaleLabel(void)
+{
+  if (my_tempscale == TEMP_CELSIUS)
+    return "C";
+  else if (my_tempscale == TEMP_FAHRENHEIT)
+    return "F";
+  else if (my_tempscale == TEMP_KELVIN)
+    return "K";
+  else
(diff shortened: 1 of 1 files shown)


universam1:
Looks good thanks !

PR #189, opened 2018-08-08
iSpindel MQTT implementation

jmelhus:
I've implemented MQTT binding for iSpindel with the use of PubSubClient library.


jmelhus:
Good, I've rebased and created a new pull request. Regarding the reference in the platformIO, sorry I forgot to add it before creating a new pull request. Any chance you can add it there? PubSubClient from knolleary: https://github.com/knolleary/pubsubclient/ Thanks for the great work and product!

PR #190, opened 2018-08-08
iSpindel MQTT implementation

jmelhus:
Based on the first pull request and your comment; I've rebased to the latest version and created a new pull request based on this. I deleted my old forked repository and forked a new one and made the changes there, sorry if there is a better way, I'm quite new to GitHub. This should be the latest version of iSpindel from you with the MQTT binding implemented from me with the use of PubSubClient library.


universam1:
Thanks for the contribution, looks fine!

jamesgol:
Too cool, I had started working on something like this a while ago but never finished it enough to submit.

Issue #96, opened 2017-09-19
Remote Configuration


DottoreTozzi:
As the topic came up again on hobbybrauer.de today (and I'm sure quite a few people would love to see a feature like this):

We should discuss what a remote configuration of the iSpindle from a local or public server might look like.

I'll implement this in the existing "local", usually Raspi-based solution first, since the Python script will also be the foundation for the future public server and a spare Raspi as an access point is useful, if not required, anyway, when using stainless fermenters as is the case in my setup, where the iSpindel's WiFi range is rather limited.

In order to keep it as simple as possible on both sides, I suggest it looks something like this:
There is a user defined interval in the iSpindel's config, like, daily, or every n sleep intervals, after which the iSpindel asks the generic TCP server for a JSON config string, mapped by hardware ID, verified by User Token (ignored in a local setup).

For example, the iSpindle opens a TCP connection and sends a JSON like this:
{ cmd: "qryCfg", ID: xxxxxxxxxxxxxx, token: "xyzxyzxyzxyz" }

The server replies with a JSON containing all relevant config values.
The iSpindle then terminates the connection, checks if there are any changes to its config (in which case it will probably reboot), else simply reopens it and transmits its data as usual.

Impact on battery life should be minimal in this scenario...
Next to nil in a local environment, where the AP and the server are one and the same.

universam1:
Good idea to discuss this topic here as to have some documentation.

I was also considering various options how to remote configure the device. 
Using JSON as the protocol is definitely the way to go and offers flexibility for future additions and extensions. 
I'm not sure yet what would be the best way to determine the update interval, as this is uptime which is directly costly battery life. It might be also remote configurable but will lead to user confusion as it would be always delayed by one iteration. Perhaps a hard coded setting could be enough to start with, or something incremental being very frequent at the beginning (new batch, everyone is nervous) and less often over the time (days, or weeks) passing by.

Though, regarding the communication protocol I feel strongly to focus on well tested implementations like REST or MQTT. On the first glance, a bare TCP connection might be easy in case of the Raspi setup but its major flaw is is not being portable at all. 

Considering a **REST** request is a well proven standard and the main benefit is its simple debugging. Regarding the python setup it is also easy to implement AFAIK. Hence, this API can be easily ported or used also with other services than the Raspi with just changing the URL.

I like the implementation of this one: https://ubidots.com/docs/api/index.html#get-a-variable and would suggest leaning towards it.

[...]

PR #201, opened 2018-09-02
adding remote config for TCP

universam1:
@DottoreTozzi here is the implementation for the TCP backend #96


FernandoGarcia:
Hi @universam1!

Here:

`interval < 24 * 60 * 60` 

I think you could make :

`interval < MAXSLEEPTIME`

Because I think that MAXSLEEPTIME was declared for this purpose.

Sorry if I'm wrong.

Best regards.

PR #202, opened 2018-09-03
Add messages to return code and limit connection attempt

FernandoGarcia:
It will prevent waste all the battery when the broker is down and add some readable messages for debug purpose.


universam1:
Good point

PR #205, opened 2018-09-07
Return "true" only if file was successfully saved.

FernandoGarcia:


FernandoGarcia:
I sent a second PR by mistake.

PR #208, opened 2018-09-09
Fix compiling errors with NO_CONSOLE

### LINKED ISSUE #206: Won't compile with NO_CONSOLE defined.
Hi!
There's compiling errors when `NO_CONSOLE` is defined.
console
Best regards.

FernandoGarcia:
Fix #206 
Also remove DEBUG variable because if user don't want console probably don't want debug messages.


universam1:
I want to remove the option the disable console as it is a support nightmare and no benefit

FernandoGarcia:
I think it can save a few milliseconds on run time.

PR #211, opened 2018-09-12
Validate input for sleep time

FernandoGarcia:
If user sends in portal page something like 0.xx the sleep time will be 0 then iSpindel won't work anymore.


FernandoGarcia:
[Here](https://github.com/universam1/iSpindel/blob/master/pio/src/iSpindel.cpp#L502) it's limited to 1 day.

PR #232, opened 2018-11-25
Native Brewersfriend support, 3 degree formula possible now

Misenkooo:
- added native support for Brewersfriend
- increased field lenght for my_token
- increased polynominal field lenght to support 3 degree formula for more precision

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -9,9 +9,23 @@
 ; http://docs.platformio.org/page/projectconf.html
 
 [platformio]
-env_default = d1_mini
+env_default = d1_mini_pro
 lib_dir = pio/lib
 src_dir = pio/src
+
+[env:d1_mini_pro]
+platform = https://github.com/platformio/platform-espressif8266.git
+board = d1_mini_pro
+framework = arduino
+monitor_speed = 115200
+upload_speed = 921600
+lib_deps =
+  ArduinoJson
+  ESP Async WebServer
+  OneWire
+  DallasTemperature
+  RunningMedian
+  PubSubClient
 
 [env:d1_mini]
 platform = espressif8266
@@ -26,7 +40,7 @@ lib_deps =
   DallasTemperature
   RunningMedian
   PubSubClient
-  
+
 [env:d1_mini_stage]
 platform = https://github.com/platformio/platform-espressif8266.git#feature/stage
 board = d1_mini
--- a/pio/lib/Globals/Globals.h
+++ b/pio/lib/Globals/Globals.h
@@ -20,7 +20,7 @@
 extern Ticker flasher;
 
 // defines go here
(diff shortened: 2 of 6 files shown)


platformio.ini
```diff
@@ -9,10 +9,24 @@
 ; http://docs.platformio.org/page/projectconf.html
 
 [platformio]
-env_default = d1_mini
+env_default = d1_mini_pro
```
universam1: we are not going to change the default here

universam1:
@Misenkooo Unfortunately, Brewersfriend is not willing to support this project although he is commercializing its use case.
This is against all interest of the community, very sad, and makes it questionable to support further

platformio.ini
Misenkooo: Sure, no problem. I have now most of ispindels with a pro board so I've changed this to default in my config.

Misenkooo:
Did you talk with them about that? I believe the support of digital hydrometers is available also in free version, isn't it?

avantassel:
> @Misenkooo Unfortunately, Brewersfriend is not willing to support this project although he is commercializing its use case.
> This is against all interest of the community, very sad, and makes it questionable to support further

How do you justify supporting Ubidots then at 8x the cost of BrewersFriend?  https://ubidots.com/pricing/

Misenkooo:
Maybe they could include iSpindel in their free membership ? Tilt and other commercial devices I would keep in paid membership only. That would be fair in my opinion... Ubidots is a generic portal where you can collect any data from any device and visualize in charts but it has no connection with homebrewing. Even in free education version you have points which are deducted from your account every day and if you are on zero you have to buy the points again. You also need to realize that collecting data from users, processing many requests also cost money because of increased hardware requirements so provide it for free is also not a way to go in my opinion. And they did not increase fees because of hydrometers support.

Also I don't think the iSpindel support is that important for a decision if to buy membership or not. I had membership before this was even announced and was happy with that and used another tools to monitor iSpindel logs. This is only nice benefit for me now.

avantassel:
> That's commercial ubibots, they also have not-for-profit free version.
> 
> Also ubidots makes no money from iSpindel support, where as BrewersFriend functionality is greatly improved by being iSpindel compatible and thus they would be making money from it

Ok you're talking about the Educational use for Ubidots that is free, I do know a lot of people at universities that use BrewersFriend, maybe they would be open to this kind of offering.

vytux-com:
That's the other problem with this pull request. 

You should always have exactly 1 fix per request

universam1:
Your effort to contribute is great and welcome. What is not welcome is the behavior of brewersfriend towards this project, and the question is do we want to encourage anyone for that platform by adding a direct support for?
I believe it should rather be discouraged and use other more brewer-friendlier services

PR #245, opened 2018-12-29
Support for Blynk Added

ldab:
Hi,

I have added support for Blynk App and also an App example, the App is free and I'll update the README as well.

I'm manufacturing a PCB to fit inside a M&M mini tube with WROOM-02 module (no LDO) for better battery life and also using 3.6V 14500 batteries.
PCB and sensors are on their way so I'm unable to test at the moment with the MPU6050.

The PCB uses Through hole components and a Ball (or mercury) tilt switch to turn it on and off (Horizontal is off) the Battery charging circuit is not protected therefore opted to make SMD as would allow "experienced" users only.
pcb_new-pcb_20181229132054
schematic_ispindel-esp-wroom-02_esp-wroom-02_20181229132130


universam1:
Sorry to say that but a 14500 is nothing different than a 18650 regarding the End of charge voltage which is 4.1-4.3V, the 3.6V is the rated voltage!
You will certainly grill the ESP that way!

ldab:
I'm aware of that, but when the device is on the horizontal, the ESP is disconnected from the battery and only then the charging circuit will be enabled.

pio/src/iSpindel.cpp
```diff
@@ -638,12 +648,105 @@ bool uploadData(uint8_t service)
     return sender.sendTCONTROL(my_server, my_port);
   }
 #endif // DATABASESYSTEM ==
-  return false;
+  
+#ifdef API_BLYNK
```
universam1: There is a sender class to handle the upload, not the main file. Clear abstraction of concerns, not hackish

ldab: I'll work on it and move to the sender class.

ldab:
I will open a clean new one

Issue #256, opened 2019-02-18
6.0.4 - cloud service parameters not saved


ploufe:
Hi,

Since I updated my device in 6.0.4. version I can no longer saved my service cloud configuration.
I'd like to use a HTTP server.
I write all config parameters, clic on "save" button, 
but after reboot my ispindle always display "Ubidot" (instead of "HTTP")...

Any idea to solve it?

universam1:
Please send a log output

ckrack:
I can confirm the problem on 6.0.2

I'm posting my log - I set the TCP server yesterday and upon first switching the device on it sends data correctly.

```log
12:32:44.040 -> FW 6.0.2
12:32:44.040 -> 3.0.0-dev(c0f7b44)
12:32:44.040 -> Worker run!
12:32:44.040 -> mounting FS... mounted!
12:32:44.040 -> reading config file
12:32:44.040 -> offsets not available
12:32:44.040 -> parsed config:
12:32:44.075 -> {"Name":"Spindel_Stern","Token":"foobar","Sleep":900,"Server":"example.com","API":6,"Port":61085,"URL":"","DB":"ispindel","Username":"","Password":"","Job":"ispindel","Instance":"000","Vfact":191.8,"TS":0,"OWpin":12,"SSID":"Foobar","PSK":"foobar","POLY":"-0.00031*tilt^2+0.557*tilt-14.054","aX":0,"aY":0,"aZ":0}
Boot-Mode: Deep-Sleep Wake
12:32:44.248 -> 
12:32:44.248 -> woken from deepsleep, normal mode
Samples:42 min:63.64 max:63.67 time:755
12:32:45.014 -> x: -9992 y: 12492 z: -1928
12:32:45.014 -> Tilt: 63.66
12:32:45.014 -> Tacc: 16.72
12:32:45.014 -> Volt: 5.34
12:32:45.048 -> Temp: 19.44
12:32:45.048 -> Gravity: 20.15
12:32:45.048 -> After waiting 201ms, result 3
12:32:45.255 -> IP: 192.168.178.40
12:32:45.255 -> 
12:32:45.255 -> calling TCP
12:32:45.255 -> {"name":"Spindel_Stern","ID":3937792,"token":"foobar","angle":63.66018,"temperature":19.4375,"temp_units":"C","battery":5.338894,"gravity":20.14841,"interval":900,"RSSI":-68}
12:32:45.322 -> Sender: TCP stream


[...]

PR #260, opened 2019-02-27
Update WiFiManagerKT.h

ckrack:
Fix incorrect API being selected.


ckrack:
Fixes #256

universam1:
Thanks for the fix

PR #263, opened 2019-03-08
Firebase Realtime Database Integration

simduchaine:
I added this Firebase Integration (on RealTime Database, but not Cloud Firestore) to push data in my database for a custom VueJS Dashboard I created.

### DIFF
--- a/pio/lib/Globals/Globals.h
+++ b/pio/lib/Globals/Globals.h
@@ -83,6 +83,7 @@ extern Ticker flasher;
 #define DTInfluxDB 8
 #define DTPrometheus 9
 #define DTMQTT 10
+#define DTFirebase 11
 
 // Number of seconds after reset during which a
 // subseqent reset will be considered a double reset.
--- a/pio/lib/Sender/Sender.cpp
+++ b/pio/lib/Sender/Sender.cpp
@@ -155,6 +155,31 @@ String SenderClass::sendTCP(String server, uint16_t port)
     return response;
 }
 
+bool SenderClass::sendFirebaseDB(String server, String url, String token)
+{
+    _jsonVariant.printTo(Serial);
+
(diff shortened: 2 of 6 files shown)


heliocosta:
Hi @simduchaine,
I'm interested in this feature as well, I did clone your branch to test it, but didn't work. I guess the issue is with me rather than your implementation. The firebase IMO is a bit confuse about what token I should use. Where did you get the token from? Can you post here how you've config the Firebase on the iSpindel UI?

Thanks,

simduchaine:
@heliocosta I kept the "token" feature as a fallback to the old Firebase API, but tokens are not used anymore in Firebase (https://firebase.google.com/docs/database/rest/auth#legacy_tokens). So you just need to leave it blank. 

Also, ensure that the url and path are entered correctly, and in your Firebase account, make sure that the database is writable (https://firebase.google.com/docs/database/security/#section-authorization)

Take a look at how I entered the urls for my corresponding database node. Server Address is for the database url and Server Url is for the node path where you want to store the data.

Annotation 2019-03-29 115218

IMG_0510

heliocosta:
I can confirm that this branch works, however, would be interesting to use the token for firebase user authentication... this way, we'd know the user that has sent that data. This would be optional.
Basically, the URL POST would be something like:

https://[DATABASE-NAME].firebaseio.com/data.json?auth_token=[FIREBASE-USER-TOKEN]

simduchaine:
@heliocosta in the actual codebase, the token is passed as a parameter in firebase functions, so it should work as described.

heliocosta:
I got it. I'll have a go and let you know if that works or not.

universam1:
sorry for the long delay, can you check if your additions works with latest source code @simduchaine ?

Issue #259, opened 2019-02-20
Blynk support not working


tdsnor:
Hi. 
Is the Blynk support released in 6.0.4? I was not able to choose this service in the drop down. But I see that Leo has been working with it.

Thanks.

Grandement:
Hi,
I am waiting for this support as well. The pull request looks really promosing. I hope we get it soon into the Master.

mr-bridger:
Is this in progress? I use blynk and would love to use that instead of ubidots.

Grandement:
Hi mr-bridge,
I have compiled the pull request of Idab myself. Worked very good. I modified his version a little bit because I like Plato more than Gravity ;). Hopefully it will be merged to the master one day.

ldab:
https://github.com/universam1/iSpindel/pull/264

Topspliff:
Hi is there any update on this? I`m also a Blynk user and this would be most excellent to have it as an option, I gave up with Ubidots as I could not work out how to use it, I`m currently using the Thingspeak option, however, I do not get the gravity reading only temp, tilt and batt showing, no idea why any thoughts most appreciated.

Cheers Alan

PR #264, opened 2019-03-30
Blynk option and email notification

ldab:
Merged commit into master. Email option for low battery.

#259

Also option to trigger calibration and configuration via App Terminal possible if we want. #TODO


Grandement:
I tried it and works well. It would be better to extend the docs with the virtual pin definition. Because I would like to reuse my token instead of copy your project with a new token. Therefore it was additional effort to look in the code for the correct pins.
The option to configure it via the terminal would be really help. This did not work for me but you already mentioned it (#TODO) 

Thanks for the awesome work.

ldab:
> 
> 
> I tried it and works well. It would be better to extend the docs with the virtual pin definition. Because I would like to reuse my token instead of copy your project with a new token. Therefore it was additional effort to look in the code for the correct pins.
> The option to configure it via the terminal would be really helpful. This did not work for me but you already mentioned it (#TODO)
> 
> Thanks for the awesome work.

Thanks for the input, once the pull request (if) is approved, I will update the docs accordingly.

ldab:
Have rebased, but don't have the HW to test it crashes without the sensors, are you able to try?

ldab:
Tested and fixed, all good :)

universam1:
@ldab would you mind to solve the merge conflicts?

ldab:
@universam1 done

pio/lib/Sender/Sender.cpp
```diff
@@ -485,3 +487,237 @@ bool SenderClass::sendTCONTROL(String server, uint16_t port)
...
+    if (_validflag != RTCVALIDFLAG)
+    {
+      Blynk.email(Subject, Content);
+      _validflag = RTCVALIDFLAG;
+      ESP.rtcUserMemoryWrite(RTCEMAILADDR, &_validflag, sizeof(_validflag));
+      delay(200);
+    }
+}
+
+uint8_t SenderClass::blynkTerminal(String token)
```
universam1: I am not in favor for another configuration procedure besides of Wifi Manager UI. It is narrow minded, possibly fitting to Blynk but not at all for general purpose. It’s a scope creep

ldab: `::blynkTerminal` ? It's not fully implemented yet on this PR... The idea behind it is to trigger the configuration without opening the container.

We can simply remove it without affecting the current functionality, please let me know.

universam1: Yes please remove it then

pio/lib/Sender/Sender.cpp
```diff
@@ -485,3 +487,237 @@ bool SenderClass::sendTCONTROL(String server, uint16_t port)
...
+    else {
+        CONSOLELN(F("\nFailed to connect to Blynk, going to sleep"));
+        return false;
+    }
+    
+    delay(300);     //delay to allow last value to be sent;
+    return true;
+}
+
+bool SenderClass::sendBlynkHTTP(String token)
```
universam1: If this http method is not used, can we remove the dead code?

pio/lib/Sender/Sender.cpp
```diff
@@ -485,3 +487,237 @@ bool SenderClass::sendTCONTROL(String server, uint16_t port)
...
+        else{
+            CONSOLELN(F("\nERROR Sender: couldnt connect"));
+            return false;
+        }
+    }
+
+    return true;
+}
+
+void SenderClass::sendBlynkEmail(String Subject, String Content)
```
universam1: Not certain about this alerting method, normally the server side is responsible for alerts not the client! Wonder what can go wrong and delay which eventually kills the battery fully

pio/lib/Sender/Sender.cpp
```diff
@@ -485,3 +487,237 @@ bool SenderClass::sendTCONTROL(String server, uint16_t port)
...
+            Blynk.virtualWrite(_pin, _value);
+        }
+    }
+
+    else {
+        CONSOLELN(F("\nFailed to connect to Blynk, going to sleep"));
+        return false;
+    }
+    
+    delay(300);     //delay to allow last value to be sent;
```
universam1: That’s obsolete, is been taken care of at the main caller

pio/lib/Sender/Sender.cpp
ldab: ok

pio/lib/Sender/Sender.cpp
ldab: ok

pio/lib/Sender/Sender.cpp
ldab: Not sure what is obsolete here, can you please clarify?

pio/lib/Sender/Sender.cpp
ldab: ok

pio/lib/Sender/Sender.cpp
universam1: I’ve implemented a connection timeout on every call to prevent hanging connection attempts. For this blynk.email I doubt a bit this is super safe, especially the condition here where we are out of battery. Hence if there are no important reasons I’d consider the api as the only communication way without another sidechannel

pio/lib/Sender/Sender.cpp
universam1: The delay here is covered already in the main routine, doesn’t need duplication

ldab: This is required for Blynk for some reason, without it the last value is not sent. I tested with 100ms and it works. So I'm afraid we need to leave it there.

pio/lib/Sender/Sender.cpp
ldab: The reason would be to warn the user about low battery, and it would only happen once. But as the brewering process goes, I would doubt someone would open the fermenter to charge the battery, so i will remove it.

Topspliff:
Are we going to see this incorporated into a new release soon?

PR #278, opened 2019-05-28
[WIP] TDS Sensor

c7h:
TDS (Total Dissolved Solids) could be another indicator of how much sugar/alcohol is still left in the brew. It might not be the best measuremet - maybe conductivity in _μS/m_ would be the better measurement. If anyone knows about this being used in traditional brewing, let me know 👍


universam1:
Interesting , could you add more information ℹ️ about this sensor and what results you experienced?

PR #286, opened 2019-07-12
Update FAQ-en.md

ErikdBr:
I edited the FAQ, hope you like it.


universam1:
Thank you!

ErikdBr:
It's a bummer it isn't working for all as you can see in the post about the offset tilt. However it solves the problem of an iSpindel where the blue led stays on permanently after first time flashing.

PR #303, opened 2019-10-06
Fix for prometheus sender data

lithorus:
Seems that after the migrating to ArduinoJSON v6 it was incorrectly
implemented in the Prometheus SenderClass. Might also affect others.


universam1:
Thank you for the fix @lithorus

PR #306, opened 2019-10-23
make InfluxDB work again (fixes #270)

### LINKED ISSUE #270: InfluxDB not working in 6.0.5
InfluxDB reporting is not working in 6.0.5. I reverted to 6.0.2 and it solved the issue. Have seen other issues reported on this e.g. #215, #218 but these are closed and fix included (I assume) in 6.0.5 so seems this is another issue or fix is not working. Don't have any info on the error, could not find any trace of connection in influxdb log and no info found in network packet capture so seems iSpindel is not even connecting to the influxdb server.

m0n5t3r:
this applies the fix from https://github.com/universam1/iSpindel/issues/270#issuecomment-535685445

### DIFF
--- a/pio/lib/Sender/Sender.cpp
+++ b/pio/lib/Sender/Sender.cpp
@@ -219,7 +219,7 @@ bool SenderClass::sendInfluxDB(String server, uint16_t port, String db, String n
     msg += name;
     msg += " ";
 
-    for (const auto &kv : _doc.to<JsonObject>())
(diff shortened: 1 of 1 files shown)


felixkosmalla:
Thank you for finding this error. I had the same problem that the iSpindel did not send any data to influxdb. This MR should be accepted.

universam1:
Thank you for the contribution

Issue #318, opened 2019-12-02
iSpindel losing connectivity when going thru Wifi extender


jmcazaux:
Hi there,

I am facing faradic issues, but while investigating, I noticed that the iSpindle seems to be losing connection when going thru a wireless extender.

Steps to reproduce:
- start the iSpindle (fully configured) near your Wifi router / AP
- check that it logs OK to your target (Ubidots, etc.)
- let it run for a while (in my case, it worked flawlessly for 8 hours at 15s interval)
- move the iSpindle close to the Wifi extender
- check that it logs OK to your target (Ubidots, etc.)
If it does not log to Ubidots, put the Spindel in config. mode (double reset) and save the config again. 
Check that it logs OK to your target (Ubidots, etc.)

What is wrong:
- at some point, Ubidots (or your target) will stop receiving data.
In my case it stopped after 20 frames.

Note : my Wifi extender is a [Netgear WN3000RP](https://www.netgear.com/home/products/networking/wifi-range-extenders/WN3000RP.aspx) with the same SSID as the main router.

I am happy to help investigating / debugging further.
I do not have experience in platform development, but I have installed the PlatformIO plugin in VSCode and with a bit of guidance I shall be able to have the thing running and logging stuff.

If this is useful, my Spindel was built from [this kit](https://www.3d-mechatronics.de/de/ispindel-teilmontiert/-partially-assembled_962.html).

Hope this helps,
JM

universam1:
Thank you for the detailed problem description.

As I read your are confirming that the iSpindel does work as expected, but there seems to be an issue with the Extender rather.

For any help here it would at least be necessary to attach the logs from the console while in this situation.

jmcazaux:
Rogare that @universam1 , I'll try to get the error captured in the console in the next days.

jmcazaux:
As agreed, here what I am consistently getting when trying to connect via my Wifi extender...

A few good frames (just one here), and then it fails on connecting to the Wifi

```
Final-sleep: 15s; RT: 5515
sl␀lܞ|␀�l�|␂␌␄␄�␌l�␄b|��␂�␓�{�#�␌#��og�lno���␌c␜p��lsd{lp�g�␘␃␄␄�␌l␄��␌␌␌c␄n�|␃l�␄␄�c��gn�␀$��l ␃�␒␓nn␄l`␂␇␃n{���g␌␄c␄�␏d␇s��o␌␌c␄�␇l�bsc␄␄��㌓`␃��g�␃
[...]

PR #320, opened 2019-12-04
Improves restoration of Wifi connection when it has been lost (#318):

jmcazaux:
- allows 50x200ms instead of 50x100ms to connect to Wifi on wake up
- when restoring the connection, wait up to 50x200ms that it is actually restored.

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -1071,12 +1071,32 @@ bool isSafeMode(float _volt)
     return false;
 }
 
-void connectBackupCredentials()
+bool connectBackupCredentials()
 {
   WiFi.disconnect();
   WiFi.begin(my_ssid.c_str(), my_psk.c_str());
-  CONSOLELN(F("Rescue Wifi credentials"));
-  delay(100);
+  CONSOLELN(F("Rescued Wifi credentials"));
+
+  CONSOLE(F("   -> waited for "));
+  unsigned long startedAt = millis();
+  // int connRes = WiFi.waitForConnectResult();
+  uint8_t wait = 0;
+  while (WiFi.status() == WL_DISCONNECTED)
+  {
+    delay(200);
+    wait++;
+    if (wait > 50)
+      break;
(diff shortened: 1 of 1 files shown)


universam1:
From my experience when a connection is not established within 5s it will hang forever. This I’ve created to prevent a battery drain. Why do you believe the 10s should make the difference?

jmcazaux:
**On restoration** (function `connectBackupCredentials`) I have witnessed a lot of occurences above 5s, averaging around 8s.
It might be the extender or my specific ISP box that creates this, but I saw a lot of occurences around 8s.

The other thing I noticed is that once the connection has failed in after a wake up it is **never** restored.
Exactly, in my setup, I never saw restoration working.

My guess is that as there was only a 200ms delay in `connectBackupCredentials` and the Spindel was going to sleep, maybe it is storing a "half broken" state that is then used on next wake up, which fails the connection again.

I am seeing a few guys in a French brewing forum that are experiencing loss of data transfer that never get restored, so I think there is something that happens around Wifi & restoration of a failed connection.

Needless to stay that it only uses battery if the connection needs more than 5s to work or be restored.
Therefore it should not impact all users that are experiencing consistent connection today, but it might solve an issue for those user who are experiencing loss of connectioon that never gets restored (indeed with a price to pay with battery life, but this is better than no data transfer).

Would you be comfortable putting this in a user configurable switch (in the config portal)?

Personally, I'd rather keep things simple and avoid confusing the user with too many options.

My 2 cents,
JM

Vadim1811:
Often power outages in the house. In case of coincidence of lack of power on an access point and time of data transfer with iSpindel, the last does not Wake up any more. How can this problem be solved programmatically, even with the loss of several transmission packets?

jmcazaux:
Hi @Vadim1811 ,
I have uploaded a binary [here](https://github.com/jmcazaux/iSpindel/releases/tag/v6.2.1-exp.1).

Please try it and you can see if it improves things for you. It certainly did for me.

PR #336, opened 2020-01-28
add two point calibration support

vitotai:
2 point calibration support. Using only 1.0 in water and OG might get decent result, linear polynomial should be supported.

### DIFF
--- a/tools/calibration/calibration.htm
+++ b/tools/calibration/calibration.htm
@@ -1,4 +1,4 @@
-<html>
+<html lang="en">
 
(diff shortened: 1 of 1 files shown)


thegreatgunbantoad:
From what I've seen on the units I have built with the CherryPhilip74 PCB, the relationship is almost always a third order polynomial.  That said I haven't gone to great efforts to keep the temperature constant. I would like to see the character limit on the input equation upped a bit though, I believe it tops out at 100.

PR #346, opened 2020-02-10
Crash in WifiManager if WIFI_MANAGER_MAX_PARAMS exceeded

pppedrillo:
Do not add params into WifiManager if WIFI_MANAGER_MAX_PARAMS exceede…d (otherwise it will cause crash)


universam1:
Thanks

PR #353, opened 2020-02-28
Added support for AWS IoT ThingShadow communication

akrto:
Hi.

Added support for using TLS on top of MQTT to connect to AWS IoT Core.
Also some changes to parsing of settings in pio/src/iSpindel.cpp to adapt to the way AWS IoT works when subscribing to delta topic.
You may not like these changes and would want to do them in a different way, please do so if you want to.

### DIFF
--- a/pio/lib/Sender/Sender.cpp
+++ b/pio/lib/Sender/Sender.cpp
@@ -11,9 +11,139 @@
 #include <ThingSpeak.h>
 #include <BlynkSimpleEsp8266.h> //https://github.com/blynkkk/blynk-library
 
+// For AWS IoT
+#include <WiFiClientSecure.h>
+#include <ArduinoJson.h>
+#include <MQTT.h>
+#include <time.h>
+
+#define emptyString String()
+
+#include "../../secrets.h"
+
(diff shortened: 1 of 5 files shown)


pio/lib/Sender/Sender.cpp
```diff
@@ -11,10 +11,140 @@
...
+time_t now;
+time_t nowish = 1510592825;
+
+// End of for AWS IoT
+
 #define UBISERVER "things.ubidots.com"
 #define BLYNKSERVER "blynk-cloud.com"
 #define CONNTIMEOUT 2000
 
+void NTPConnect(void)
```
universam1, line 65: The battery life concept is "fire and forget" as every second runtime significantly reduces battery life time. Therefor I avoid lookups by all means which eventually cause delays or potentially even hang.

akrto: Yes, in TLS, verifying certificate validity is optional, and since we normally settle for regular http, I guess security is not as much an issue, as it is something to "get past".

I´ll have a look at scrapping time lookup, I have actually had a hangup during a brew recently, switched to plain old http after that.

PR #385, opened 2020-06-14
Increase DynamicJsonDocument to size*3

gingerneil:
Increase DynamicJsonDocument to size*3
Shown to overcome 6.2.0 saving issues with some gyro boards


universam1:
Thanks @gingernail

PR #386, opened 2020-06-14
Temperature Type Selection Update

thegreatgunbantoad:
When selecting the temp units in the in the configuration, they only seem to change if you select twice before saving. Moving the sTS; call out of the window.onload in to a window.onchange function seems to resolve it.


universam1:
@thegreatgunbantoad thank you

PR #392, opened 2020-06-17
Update calibrate.py

apro:
3rd degree poly - also works with iSpindel


universam1:
Thanks for contribution @apro

PR #412, opened 2020-08-14
calibration.htm outputs polynomials that can be copied directly.

vitotai:
To output polynomials that can be used directly. Using "tilt" instead of "x", so that the polynomials can be copied without modification.


universam1:
Excellent @vitotai thank you for that!

universam1 (approved):

PR #415, opened 2020-08-23
Fix url and uri mixing

isaul32:


universam1 (commented):
the html id `url` is changed to `uri` does that catch the identifier?

isaul32:
It has changed uri to url in here https://github.com/universam1/iSpindel/commit/6d1219fd74d8893fdf7e1474443c0ad0ca01cb18#diff-46b0b7d03759ea011d1640d980af8453L44-L55 but uri is still used in iSpindel.cpp. The bug adds useless url/uri input to configuration menu (except generic HTTP API which actually use it).

universam1:
Indeed excellent research, thank you!

universam1 (approved):

PR #422, opened 2020-09-10
MQTT improvements

Radu-nic:
Hi guys - this is my first contribution ever on GutHub ;) so ... be gentle with me :D. **I love this project!!! Great job!**

I have used mqtt connection and i've found that something must be fix and other things should be nice to have so ... 

 MQTT improvements:
 - may change settings live, reading a persistent mqtt message with the same structure as the one saved in FS. The topic read is "ispindel/<<Name>>/settings".  Every option saved into persistent local settings may be changed, but POLY and Sleep was those of maximum interest. After reading this new settings message this will be erase.
  - sent to mqtt broker an json report containing all readings on report topic This was useful for future writing to a log file or in my case to a database

MQTT fix:
  - fix on sending mqtt messages without user/password. If you do not set a username or password some mqtt brokers don't accept the message sending  empty user&password
  - do not try send message if not connected after 3 retries.

### DIFF
--- a/pio/lib/Sender/Sender.cpp
+++ b/pio/lib/Sender/Sender.cpp
@@ -39,6 +39,12 @@ void SenderClass::stopclient()
     delay(100); // allow gracefull session close
 }
 
+void SenderClass::registerMqttCallback(MessageCallback callbackProc, bool clearReceivedPersistentMessage)
+{
+    ClearMqttPersistentMessage = clearReceivedPersistentMessage;
+    messageCallbackProc = callbackProc;
+}
+
 bool SenderClass::sendMQTT(String server, uint16_t port, String username, String password, String name)
 {
     _mqttClient.setClient(_client);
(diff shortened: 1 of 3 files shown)


universam1 (changes requested):
Nice work!
I’d like the remote configuration feature being optional  which is disabled by default.
 Reason it that it has a direct penalty on the runtime hence battery lifetime and that is why someone must be aware of that when enabling this option. 
Sorry for the delay @Radu-nic there were conflicting PR arriving

Radu-nic:
Hi Samuel.  
I'm glad you had time to take a look at those changes and that you liked my work. I saw you fixed the issue sending mqtt message without connection, great! You do have right indeed about extra processing on "settings change" message. I will change the code to trigger this only when needed. Also I am trying to update my fork to master or cherry-pick your commits but ... no success. I will revert my branch and rewrite the changes. 
So i suggest we abandon (delete) this pull request for the moment, don't know how to do it, i'm sure you do ;).
"I'll be back"

PR #431, opened 2020-11-12
Clean up following item #410

thegreatgunbantoad:
Ensure all tilt calcs are inline with #410
Remove unused pitch and rolls variables
Upversion firmware ID


ErikdBr (commented):
Coding looks good.

universam1 (approved):

PR #434, opened 2020-11-17
Update Sender.cpp

msbannert:
FHEM upload does not work. Looks like i am the only user of it ;-)


thegreatgunbantoad (commented):
Having never used FHEM I can't attest to this working.  But it looks pretty reasonable and it compiles fine.

universam1 (approved):

universam1:
thanks

PR #439, opened 2020-12-03
Temperature Adjustment Feature

thegreatgunbantoad:
Allows the raw sensor value to be adjusted based on the calculation in the Config screen of the AP.
Calculation is done on the scaled version of temp so that +1 would add 1 decC, 1 degF or 1degK based on what the unit is setup for.


thegreatgunbantoad:
Just noticed a bit of other rogue code.  Will fix and generate a new pull request

PR #440, opened 2020-12-03
Temperature Adjustment Feature

thegreatgunbantoad:
Allows the raw sensor value to be adjusted based on the calculation in the Config screen of the AP.
Calculation is done on the scaled version of temp so that +1 would add 1 decC, 1 degF or 1degK based on what the unit is setup for.


thegreatgunbantoad:
@universam1 just a quick poke

thegreatgunbantoad:
@ErikdBr are you up to giving this a review?  Don't know how your C++ is.

ErikdBr:
> @ErikdBr are you up to giving this a review? Don't know how your C++ is.

I can read it, will take a look at it tomorrow, as it is 23.00 hours overhere.

thegreatgunbantoad:
> > @ErikdBr are you up to giving this a review? Don't know how your C++ is.
> 
> I can read it, will take a look at it tomorrow, as it is 23.00 hours overhere.

Cheers.

ErikdBr:
> > > @ErikdBr are you up to giving this a review? Don't know how your C++ is.
> > 
> > 
> > I can read it, will take a look at it tomorrow, as it is 23.00 hours overhere.
> 
> Cheers.

Have to postpone to tomorrow, sorry.

ErikdBr (changes requested):
I reviewed the code and didn't see any faults. However I am not able tot test it.

thegreatgunbantoad:
> I reviewed the code and didn't see any faults. However I am not able tot test it.

Thank you for looking.

the code has worked fine for me.

Just need @universam1 to decide if he want's to take the code or not.

thegreatgunbantoad:
@universam1 any thoughts on weather or not you want to incorporate this?

universam1 (changes requested):
@thegreatgunbantoad  I'm not sure why this is necessary.

The `temp` is already available as variable in the gravity caculation:

https://github.com/universam1/iSpindel/blob/master/pio/src/iSpindel.cpp#L1065

Hence someone can just simply put following expression:

```
x*tilt^2+y*tilt-z + temp*a^2 + temp*b ....
```

thegreatgunbantoad:
@universam1 This is to allow the temp sensor reading to be adjusted and essentially calibrated.  Nothing to do with Gravity really.

pppedrillo:
DS18B20 sensor has +-0.5c accuracy and does not require any sort of calibrations.
http://datasheets.maximintegrated.com/en/ds/DS18B20.pdf

Capture

thegreatgunbantoad:
@pppedrillo some of the ones off eBay are clones I think and they do vary a bit compared to the real thing.

pppedrillo:
@thegreatgunbantoad Ahh, I see. Then it make sense. While chasing all the ghosts from aliexpress (dodgy gyros, wemoses, temp probes etc) with software fixes may be a tedious task.

universam1:
I also second that point of @pppedrillo  that they are calibrated by manufacturers and I doubt the majority has more precise sensors available at home. 

In case if clones exist then there are other issues as well like communication problems, unexpected ids and so forth. Hence I’d rather add a note to be alert on such cheap clones sand avoid them by all means, because we cannot cover for all of their potential problems and we do not want to encourage that counterfeit business model as well

thegreatgunbantoad:
There's so little way of telling if they are counterfeit these days.  I believe the ones I have to be real, but I do see more variability in them than I'd like.  Left in air my thermocouple reads 19.5 C and two iSpindels report 21.5  and 21.9.  That's a full 2 C out.  Then again I have one in a fermenter that's at 17.6 (Inkbird and the thermocouple agree on that) reading 17.9 which is very close.

universam1:
The influence I cannot stress enough is the self heating during continuous run of the ESP. 

Always take the temperature with a grain of salt while in AP mode, don't underestimate the influence of MCU heat dissipation.

It will only be accurate when in normal mode, when the intervals are longer than ~5min.
Cant be stressed enough.

pppedrillo:
> The influence I cannot stress enough is the self heating during continuous run of the ESP.

I agree with @universam1 on this. Especially if you have your test ispindels inside closed vials. 
Are they the same amount off if you take the assemblies out of vials?

thegreatgunbantoad:
Running in normal mode at 15mins reporting to BF, one in a vial one out and temperature soaked for an hour.  All put close together, no particular draughts in the room.

PR #449, opened 2020-12-24
Patch for Blynk API for better accuracy

MatthBou:
A problem happens when data are sending to Blynk and using the specific gravity.

The specific gravity value is rounded only two digits after the decimal point, which takes away from the precision.
for example :

specific gravity read by the iSpindel: 1.052
specific gravity sent to blynk: 1.05

Thank you for the work done, this project is really awesome!

### DIFF
--- a/pio/lib/Globals/Globals.h
+++ b/pio/lib/Globals/Globals.h
@@ -20,7 +20,7 @@
 extern Ticker flasher;
 
 // defines go here
-#define FIRMWAREVERSION "6.4.1"
+#define FIRMWAREVERSION "6.5.0"
 
 #define API_FHEM true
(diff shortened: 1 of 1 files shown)


thegreatgunbantoad (approved):
Whilst I can't check that this works using Blynk, it's a pretty simple change and it certainly compiles.
Looks good to me.

gdolling (approved):
I have tested this change with blynk. Works as expected and is a good improvement. Gravity really needs 3dp.

image_6483441 (1)

MatthBou:
@gdolling  thank's for your feedback!

universam1 (approved):
:+1:

PR #451, opened 2021-01-05
Commit to provide secureMQTT with X.509 certs as an option in sender class

camsaway:
This is pre-cursor work to AWS IoT Core integration option that requires secure connection.

The main features here are:
1) Creation of a method, SenderClass::SendSecureMQTT which is a copy of SendMQTT class with parameters for secure connections
2) Modification of SenderClass::MQTTConnect to select WifiClient or WifiClientSecure object based on input parameters
3) Due to memory constraint I had to reduce JSON buffer to 256 from 1024 in sender.h. This seems to be sufficient from checks I did on ArduinoJSON test site for the standard messages we send.

I chose this approach to avoid changes to other areas of the code.

I will submit a seperate pull request with changes to Globals, iSpindel.cpp and WifiManager to enable AWS as a connection option

NB: This is my first github contribution and I'm not a developer so please excuse (and explain) any process mistakes I've made.


pio/src/iSpindel.cpp
```diff
@@ -335,20 +336,20 @@ bool startConfiguration()
   WiFiManagerParameter custom_api("selAPI", "selAPI", String(my_api).c_str(),
                                   20, TYPE_HIDDEN, WFM_NO_LABEL);
 
-  WiFiManagerParameter custom_name("name", "iSpindel Name", htmlencode(my_name).c_str(),
+  WiFiManagerParameter custom_name("name", "iSpindel Name (AWS ThingName)", htmlencode(my_name).c_str(),
```
universam1, line 339: AWSIOT should not be that prominent as it would be the suggested endpoint.

In fact rather I'd put a warning that the SSL & NTP will add a time penalty which will negatively impact battery lifetime

universam1 (changes requested):
I like this implementation, if we can remove the too prominent AWS customization of the UI I'm happy to merge

pio/src/iSpindel.cpp
camsaway: Hi, I'm happy to make those changes. How do I do it? Do I change on my branch and submit a new pull request? Or is there a way to do here?

One note though, I've spent quite a bit of time and have been unable to replace secrets.h wih input parameters on the WIFI Manager. Is this viable base code if users have to edit a secrets file and compile for themselves? (Or any suggestions on how to capture these through WiFiManager?)

LinusU: @camsaway if you push a new commit to your branch (`AWS`) it will show up in this pull request

camsaway: Thanks Linus. Will let you know when tested.

camsaway: Hi @universam1 - I have made changes as requested. I removed reference to AWS from the main GUI and if you select AWSIOTMQTT as the API then you get a warning message about battery usage as well as the "AWS Hints".
I also added a Readme_AWS.md file

camsaway:
Hi @universam1 - I have resolved the UI issue as requested. Can you please review again

gdolling:
I was initially confused by the box with "<<<< >>>>" in it and thought the thingname was to go in there. I would recommend changing the text to "name above must be thingname". It's only a minor point though.

After that I got it to work with no problems. 

I like this code and this solution. AWS IOT is not for the inexperienced though and considering the level of competency required with policies, certificates, etc to get data to appear in AWS at all. Then consider even at that point you won't get any graphs or even have the data stored until you configure a whole host of other AWS services. I think this code should be in master, but as it is such that your average user is pushed towards ubidots or blynk solutions. Given the competence required to use AWS I don't think having to compile the source is any real barrier, but does help steer the unwary away from it.

camsaway:
Hi @universam1 are you happy with the changes or are there more needed?

universam1:
> Hi @universam1 are you happy with the changes or are there more needed?

Thank you @camsaway for applying the suggestions - looks great to go!

universam1 (approved):

PR #453, opened 2021-01-08
Increased the maximum size of the polynomial parameter

niki77:
Increased the maximum size of the polynomial parameter from 100 to 250 char

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -75,7 +75,7 @@ char my_username[TKIDSIZE];
 char my_password[TKIDSIZE];
 char my_job[TKIDSIZE] = "ispindel";
 char my_instance[TKIDSIZE] = "000";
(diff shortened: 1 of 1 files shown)


ErikdBr (changes requested):
I reviewed the change, it looks ok.

ErikdBr:
What usually seems to be forgotten is to increase line 23 #define FIRMWAREVERSION "6.5.0" in the globals.h
or is that only to be done by @universam1 ?

thegreatgunbantoad:
I'd be inclined to change the 250 in:
`WiFiManagerParameter custom_polynom("POLYN", "Polynominal", htmlencode(my_polynominal).c_str(), 250 * 2, WFM_NO_LABEL);`
to
`sizeof(my_polynominal)
So as to reduce the amount of hard coding

But it's functionally fine as is.

universam1 (approved):
:+1:

Issue #436, opened 2020-12-01
Temperature Compensation, any one interested?


thegreatgunbantoad:
## I'm submitting a ...
  - [ ] feature request

## Features

I have temperature compensation working based on the calcs in the HTML poly generator doc.  It doesn't do a lot, a couple of SG points if your at 30deg but calibrated at 20deg for example.  It needs some polish to add it to the main project, I've never bothered as I don't really care for it, just did it for the challenge.  Does anyone think it's worth adding, I might get my finger out and polish it up a bit.

quarky42:
I would really appreciate this being added.

Personally I find error to be more of a concern in the situation where sources of error are additive. 

Being able to toggle this compensation on or off, with it being off by default, might be ideal.   

Alternatively, having both the uncompensated and the temperature compensated value would let someone decide which value they want to use.   If I know there is CO2 present in the must, I'll go with the uncompensated value because the dissolved CO2 can throw the reading off quite a bit anyway.   

At the end of a batch when the yeast activity is tiny or non-existent, there should be an ever decreasing amount of CO2 present.   Also, the yeast aren't generating their own warmth so one of the temperature variables is simplified.   At that point I would like to see the compensated value.

So, yeah I see a use case for having both pieces of information available.

thegreatgunbantoad:
Sorry for the slow response, been working on getting a pressure senor to work.  Nearly there (code is still a bit unstable), just need to figure out how to pass it to BrewFather and get it so that you can select your units.

I'll probably work on cleaning the temp comp up over my Christmas break.

quarky42:
No big rush, just a very nice to have thing.  Thank you for adding this.   My initial thought would be to have the existing variable output as it is current.   Perhaps a new variable can be output as "gravity-corrected" or "gravity_corrected", something like that.  If it needs to be something short then:  gravitycorrect or gravitycor with or without hyphen or underscore.

MonTage69:
Noticed that when cold crashing the gravity goes up. 
20C = 1.011
2.5C = 1.020

Maybe this is purely on the iSpindel side.

Pretty sure my calibration was done correctly as it's always been pretty spot on on my manual readings within 0.001 of a point.

I know the tilt is able to adjust. 

Any pointers?

quarky42:
[...]

PR #457, opened 2021-01-24
Automatic Temperature Compensation (ATC) feature for gravity readings

pppedrillo:
Hello.

These changes introduce Automatic Temperature Compensation (ATC) for gravity readings for iSpindel.
This ATC feature is optional and can be turned on/off and configured via portal configuration page.
ATC calculations can be configured to use either internal or custom formula.

In "Internal mode" user has to configure gravity units he is using in gravity polynomial and calibration temperature. After that software will automatically adjust gravity readings using the same formula like in iSpindel calibration Excel sheet. 
For all calculations software will respect temperature units (F or C), converting units as it needs.

For calculations using "Custom" ATC formula user has to provide his own compensation formula.
In this mode software will never do any conversions for units (eg never automatically converts C to F, F to C, SG to Plato, Plato to SG etc).
Output gravity will be calculated according to input:
1. Custom ATC polynomial, which accept as inputs a current temperature, calibration temperature and gravity already calculated using gravity polynomial (pretty much the same way as polynomial for gravity does with tilt and temp)
2. Calibration temperature

And cherry on top:
3. A math operation what has to be done with calculated adjusted gravity: 
a) calculated adjusted gravity can be used "as is" (which is 99.999% use cases I believe)
OR (those 2 options for real geeks)
b) calculated adjusted gravity can be added to "unadjusted" gravity readings
OR
c) calculated adjusted gravity can be multiplied by "unadjusted" gravity readings

Cheers and have fun :)


MonTage69:
Thanks @pppedrillo will have to give it a try on my next batch.

gdolling:
I tested the following :
Enabling/disabling works as expected. 
Forcing the temp to 20, gives identical SG readings when enabled.
Forcing the temp to 10 and 30 gives adjustments that look consistent. (SG goes down and up respectively).
It's well done, in the sense it makes no difference until enabled so will be safe.
It seems a good improvement, given that temp is readily available it seems a shame not to compensate for it. 

I did not try it with fahrenheit, nor did I try my own formula. I think I should be able to validate the actual numbers using the spreadsheet. I plan to try this in the future.

pppedrillo:
> I tested the following :

@gdolling Thanks alot for testing! Very appreciated.

universam1 (changes requested):
@pppedrillo I'm not sure why this is necessary.

The `temp` is already available as variable in the gravity caculation:

https://github.com/universam1/iSpindel/blob/master/pio/src/iSpindel.cpp#L1065

Hence someone can just simply put following expression:

```
x*tilt^2+y*tilt-z + temp*a^2 + temp*b ....
```

thegreatgunbantoad:
@universam1 The poly would get insane if you tried it that way. You'd be jamming a poly into a poly and it's limited to 100 chars.

ErikdBr:
> @universam1 The poly would get insane if you tried it that way. You'd be jamming a poly into a poly and it's limited to 100 chars.

It is actually 250 now as per pull request "Increased the maximum size of the polynomial parameter #453" which is merged today.

thegreatgunbantoad:
Looking again it's actually one poly multiplied by another. You're still looking at a poly that would be horrific, but yes with 250 chars it is now possible.  You'd also end up having to build a poly that is temperature base specific which isn't all that desirable.  How valuable this all is given that all these floaty devices aren't ever crazy accurate is debateable.  But if you are going to do it, I think coded is preferable.  I have my own firmware build which does the same thing never got round to formalising it.  I don't really use it anyway.

pppedrillo:
@universam1 You right,  it is possible to use temp in gravity poly , but there some flaws-

- as @thegreatgunbantoad said - even having 250 (500, 800 etc :) ) chars poly gets insane
- any possible input errors break the gravity calculation to void
- the ATC poly from calibration HTML already has more than 200 symbols even if var names will be 1 letter only
- the ATC poly requires alot of correct conversions from one units to another (C to F, SG plato and back if required)
- as you noticed looks like vast majority are not actually using temperature in gravity poly (I suspect because of above)
- it might be so that the way user can configure ATC in UI is more user friendly and easier to understand for them
- etc etc etc :)

If you allow, my humble proposition would be rather to completely remove "temp" var from the "main" gravity calculations, and consider gravity calculations result as a  "standard reference" gravity (like eg glass hydrometers do - they are precise at only specific temperature). All the adjustments to this gravity readings (ATC or anything else if need ) can be done after "standard reference gravity" calculated. If it is ok - I can make these changes. But its up to you.

Just my 2 cents. Cheers! :)

thegreatgunbantoad:
Turns out it's not as vile as it could be:
Assuming a reference temp of 20 C (68 F) the compensation poly simplifies to:
Deg C:  ((-1.3567E-8*(temp^3))+(5.8822E-6*(temp^2))+(-2.0288E-5*temp)+(9.9816E-1))
Deg F:   ((-2.3262E-9*(temp^3))+(2.0388E-6*(temp^2))+(-1.3461E-4*temp)+(1.0005E0))
So you just multiply your normal one by that in the poly field.

An example for SG with a 20deg ref (148 chars):
((1.0000E-6*(tilt^3))+(-1.3137E-4*(tilt^2))+(6.9680E-3*tilt)+(9.0704E-1))*((-1.3567E-8*(temp^3))+(5.8822E-6*(temp^2))+(-2.0288E-5*temp)+(9.9816E-1))

See (This needs the absolute bleeding edge version of Excel btw):
[TempComp.xlsx](https://github.com/universam1/iSpindel/files/5913075/TempComp.xlsx)

pppedrillo:
@thegreatgunbantoad  Right, there are usually many ways to achieve the same results. 
Many brewers just not a math geeks at all, thus may prefer simply tick a checkbox, choose their gravity units and press "Save".
While ofc some of them eager to invent their own formulas , transfer crazy numbers between different applications, try and fail, redo and repeat etc etc
Every1 here goes his own way as everywhere else.

ErikdBr:
As I saw this for the 1st time I thought: yes offcourse we need this. But along the way and seeing the debate above I started to doubt.

As it is allready possible to use temperature compensation in the poly especially with the increased size of the poly,  I personally see this as in between "could have" and "won't have" of the MoSCoW principle of software development.

In my experience with homebrewers for which I have build an iSpindel, most if not all of them use it to see if fermentation is ready or not from their warm and comfortable couch instead of running to their shed thru snow and rain to watch the airlock for a while to see if and if so how many bubbles per minute are coming thru. And if it has stopped bubbling, wondering "is my fermentation ready or should I let it go for another day?"
If using a blow off, most of the times there isn't any activity to see so when is fermentation ready, should I open and measure or wait another day?

So many brewers seek reassurance and confirmation opposed to an as exact as possible SG.
Measurements of OG and FG should be done by hand anyway.

So my opinion is to keep the code as lean and simple as possible to serve the majority and not implement functionality for a minority.

This is my opinion and in no way meant to be an attack of some sort to anyone or their much appreciated work. I just am thinking out loud.

thegreatgunbantoad:
@ErikdBr I'm inclined to agree with you there.  Also, these floaty things are inherently inaccurate.  Best as a guide to a flattening off of the curve.

thegreatgunbantoad:
[TempCompV2.xlsx](https://github.com/universam1/iSpindel/files/5915276/TempCompV2.xlsx)
Simpler and more platform independent form of the multiplier for temp comp assuming I got he underlying correct.

pppedrillo:
@

> As it is allready possible to use temperature compensation in the poly especially with the increased size of the poly, I personally see this as in between "could have" and "won't have" of the MoSCoW principle of software development.

@ErikdBr Have you or any1 of whom you know EVER use temperature compensation extra poly directly on gravity poly? No? Why? :)

> I'm inclined to agree with you there. Also, these floaty things are inherently inaccurate. Best as a guide to a flattening off of the curve.

@thegreatgunbantoad  Now you are seem contradicting yourself :)
 https://github.com/universam1/iSpindel/issues/436
https://github.com/universam1/iSpindel/issues/436#issuecomment-767140118

2all discussion participants: First of all I'm actually happy that we have this discussion and mot offended at all (no worries, @ErikdBr :)). 
Then... As I already said there are pros and cons. Always. Many of my friends and myself personally found this implementation much more handy than struggling with one dreaded poly. Your mileage may vary.
I just shared what I have already done anyway.

ErikdBr:
> @ErikdBr Have you or any1 of whom you know EVER use temperature compensation extra poly directly on gravity poly? No? Why? :)

No I don't know anyone who've used it because they don't see the need to, and/or probably also because it is complicated enough without it :-) and/or because it is not as easy as "check a box". 

That and all of the above said, I can imagine there are people who really want this because they allways cold crash or brew beers that ferment at much lower temperatures. And tbh apart from the major code update it's really neat to check a box and have it taken care of for you.

So I change my opinion and say let's do this and yes maybe I even will start using it once it is there :-)

thegreatgunbantoad:
Until the poly is updated any non coders won't be able to try smashing one huge poly together.


For my own curiosity, with this section of code:
```
double_t ATCModule::applyCustomFormula(double_t measured_gravity)
{
    double_t corrected_gravity = measured_gravity;

    corrected_gravity = calcCorrectedSGCustom(corrected_gravity);

    return corrected_gravity;
}
```
Is there any particular reason for the local variable and not just:
```
double_t ATCModule::applyCustomFormula(double_t measured_gravity)
{
    return = calcCorrectedSGCustom(measured_gravity);
}
```
At which point why is a function just wrapped in another function like that.  I assume there's some coding reason behind it.

thegreatgunbantoad:
For anyone wondering who might not have the skills to compile and load this the interface looks like this:
Screenshot_20210203-184910 croppedpng
It's handled really nicely and hides itself away if the check box is unchecked, which is pretty cool.

@pppedrillo A couple of points as I've now had a quick look at it running and some of the code.
Your firmware version is still reporting as 6.5.0 which it really shouldn't be (Globals.h line 23).
The ATCModule.cpp and .h files may be better placed in the pio>lib folder inside their own folder in keeping with the general style things have been done in so far.
There are also some commented out blocks of code that can probably be stripped out now or popped in a #ifdef block.

pppedrillo:
@thegreatgunbantoad 
Firmware version usually must be changed on the master branch before merging the feature. At this moment no one can predict the future version number.
File location is discussable, I have no any preferences for it and can move files anywhere agreed, np.
Commented blocks will still there. It will not hurt any1. Those are only debug console output for quick check. IMO it is very convenient to simply uncomment em for debugging rather than re-type everything from scratch every time. 

And **thanks** for pointing into leftovers - in version like it is now, function call indeed must be simplified. As said, those are just leftovers from the time I wasn't sure about if the measured gravity has to be transformed somehow before applying a poly. I pushed the changes, they are trivial.

pppedrillo:
Because of no interest to have this feature at all but rather the only verbal buzz about it, I close my pull request.
Have fun with input of 256 char polynomial and tweaking excel files.

Maclav1:
Nooo.. I was hoping this would be in the master branch by the time my next beer was ready to go. Ah well, guess I'll see if I can compile my own with you excellent changes.

ChrisThomas42:
Please can we have temperature compensation, pretty please?  As has been said these wonderful devices aren't the most accurate, but I'd rather not see the SG graph go up and down as the temperaure changes if the capabilty is available.

Alneyt:
@pppedrillo I am really sad to read "Because of no interest to have this feature at all,". This function is going to be the most useful function for me since I am making mainly lager/cold crash/diacetyl rest etc. I wish you to continue and will be very much appreciated if you complete it. thanks in advance

pppedrillo:
@Alneyt @ChrisThomas42 That was exactly my thoughts about usability and convenience of this feature when I was implementing it.
However owner of the project stated that fiddling with "temp" variable in "main" gravity poly is more than enough.

Maclav1:
I built ver 6.5.1 with ATC.
Caveats: I am in no way responsible if you brick your spindle. This is a random binary file from a complete stranger on the internet and is probably full of viruses and ransom ware and will likely eat your children, cause your car to break and ruin your beer. I am sure this breaks all kinds if github tos too.. but I am really not interested in forking the project and dealing with the maintenance and splitting the community and all that jaz.

https://drive.google.com/file/d/1LpUIe2XNduj2V_tDMuaiA5QmEuh2mFn-/view?usp=sharing

ChrisThomas42:
I'm not sure if I did wrong but I posted about this feature on UK and USA homebrew forums, asking if indeed there is any interest?

ErikdBr:
@pppedrillo Why don't you reopen this. What @Maclav1 is doing is not the way to go.

pppedrillo:
@ErikdBr After some time Stale Bot will close this request automatically anyway if here no go from @universam1

thegreatgunbantoad:
@ErikdBr it might not be the preferred option but the uprising wants what it it wants.  Reopening will do nothing @universam1 has shown no desire to accept this.
Yes it can be done through some very awkward poly, but that option isn't user friendly at all.
@pppedrillo did some great work. I have my own opinions of some tidy up, but that doesn't mean I don't think it's worthwhile overall.

Perhaps I am reading @universam1 s general response though.

Also we don't know what @universam1 s current situation is or where his head space it at, so have to reserve judgement to some extent.  It is frustrating though to have such a great project and not feel like we have a voice to influence it.  At the end of the day we are all here because we want it to be the best it can.

ChrisThomas42:
I know this is closed, just adding a comment.  From the post I put in UK and USA forums there were only a handful of positive replies, so I guess nobody wants this feature after all.  Very sad!

Issue #460, opened 2021-01-30
Change WiFi name


kitmir:
Hi,
I would like to calibrate multiple iSpindel at the same time.
The problem I face is that all iSpindels have same wifi name and therefore it is not possible to know which one I am connecting.
Only option I see is that I need to open the lid and switch the one I do not use.
Is there anyway I can change the Wifi name from the menu?
Thanks!

ErikdBr:
The WIFI name of the iSpindel in configuration mode is allways the same and can't be changed.
So you have to calibrate them one by one.

JochenV1976:
Shouldn't the WiFi name be unique for every iSpindel? Somewhere in the doc it's written that the WiFi name should be iSpindel_xxxxxx?

thormj:
It is now... we can probably close this for a bit.
I have a patch that puts the "Name" so if it's not been configured it's the node ID, but if it is, it becomes "iSpindel_White" .

PR #462, opened 2021-02-07
Add ChipId to ssid

MarcoCLA:
Generate unique ssid for each iSpindel to be able to configure them simultaneously. 
See also #460

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -394,7 +394,9 @@ bool startConfiguration()
   wifiManager.setConfPSK(htmlencode(my_psk));
 
   CONSOLELN(F("started Portal"));
-  wifiManager.startConfigPortal("iSpindel");
+  char ssid[16]; //Exactly match size "iSpindel_123456\0"
+  snprintf(ssid, sizeof(ssid), "iSpindel_%06X", ESP.getChipId());
+  wifiManager.startConfigPortal(ssid);
 
   strcpy(my_polynominal, custom_polynom.getValue());
 
--- a/docs/FAQ-en.md
+++ b/docs/FAQ-en.md
@@ -9,7 +9,7 @@ The best way to find out what the iSpindle is doing  is to read its log. This ca
 Here is a youtube how-to (in German): https://youtu.be/6xDW5GbIbUw
 
 - ***How get to the iSpindel - the portal***  
-Switch it on, press the reset button 3-4 times which sets up an access point. That means you will see a new Wifi network "iSpindel" where you connect to with your computer, smartphone or tablet. Now the browser should automatically redirect you to the page [http://192.168.4.1](http://192.168.4.1). If not, navigate there yourself.
+Switch it on, press the reset button 3-4 times which sets up an access point. That means you will see a new Wifi network "iSpindel_######" where you connect to with your computer, smartphone or tablet. Now the browser should automatically redirect you to the page [http://192.168.4.1](http://192.168.4.1). If not, navigate there yourself.
 If you are now on the portal page, under `Configuration` you can enter your token and the wifi password that you need for your network. Then click on `save` and it will send the data to Ubidots. This means that the **iSpindel is not reachable in normal operation**, until you manually start the portal again.
 
 - ***Why can't I access the iSpindle in normal operation***  
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -8,7 +8,7 @@ Am Besten du liest mit was die Spindel als Informationen ausgiebt. Das geht übe
 Ein Youtube HowTo: https://youtu.be/6xDW5GbIbUw
 
 - ***Wie komme ich auf die iSpindel - das Portal***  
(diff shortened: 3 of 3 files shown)


camsaway:
Looks like a simple fix to your problem.

How do you identify which iSpindel is which ChipID? Is it stamped/stenciled? Or do you have to work this out first? If not, is there any marked code that can be used?

MarcoCLA:
```ESP.getChipId() returns the ESP8266 chip ID as a 32-bit integer```
Which are also used as the last 4 bytes of the mac-address. So they are not identifiable for the markings. But each iSpindel will have a unique (with a very low chance on collisions) wifi name.

**Comment by pppedrillo** [comment]
@MarcoCLA  I believe, @camsaway meant the following situation: you and 10 iSpindels floating in the jar in front of you. Indeed with that change you can connect to each of iSpindel separately, but how can you know which one of those 10 floating is iSpindel_WXYZ22 and which one iSpindel_18ABCD if they are not marked?

**Comment by ErikdBr** [comment]
> .....but how can you know which one of those 10 floating is iSpindel_WXYZ22 and which one iSpindel_18ABCD if they are not marked visually in advance somehow?

And if you do know, I don't think you can connect to all of them at the same time on one device.

**Comment by pppedrillo** [comment]
> And if you do know, I don't think you can connect to all of them at the same time on one device.

Of course you can't. But you can connect to all of them 1-by-1 and get the readings without need to turn them (all but "current" one) on and off.

**Comment by thegreatgunbantoad** [comment]
@MarcoCLA quite a nice update

File: pio/src/iSpindel.cpp [hunk 1]
```diff
@@ -324,7 +324,7 @@ String htmlencode(String str)
 
 bool startConfiguration()
 {
-
+  String autoconf_ssid = "iSpindel_"+String(ESP.getChipId());
```
universam1, line 327: We can omit this memory allocation when we make the string literal

universam1 (changes requested):
Small request to omit the variable, otherwise great idea!

MarcoCLA:
I'll update the request tomorrow

pio/src/iSpindel.cpp
```diff
@@ -393,8 +392,11 @@ bool startConfiguration()
   wifiManager.setConfSSID(htmlencode(my_ssid));
   wifiManager.setConfPSK(htmlencode(my_psk));
 
+  char my_ap_ssid[20];
+  sprintf(my_ap_ssid, "iSpindel_%06X",ESP.getChipId());
+
   CONSOLELN(F("started Portal"));
```
pppedrillo, line 398: This is essentially the same as your first version, just a bit more awkward.

pio/src/iSpindel.cpp
```diff
@@ -393,8 +392,11 @@ bool startConfiguration()
...
-  wifiManager.startConfigPortal(autoconf_ssid.c_str());
+  wifiManager.startConfigPortal(my_ap_ssid);
```
pppedrillo, line 400: wifiManager.startConfigPortal("iSpindel_"+String(ESP.getChipId()));

MarcoCLA: This will return a StringSumHelper and not a char*

universam1: made a couple of tests, and while the diff is not a lot the inline String is the cheapest one:
```suggestion
  wifiManager.startConfigPortal((String("iSpindel_") + ESP.getChipId()).c_str());
```

> String literal
> RAM:   [=====     ]  47.7% (used 39092 bytes from 81920 bytes)
> Flash: [====      ]  44.1% (used 461032 bytes from 1044464 bytes)
> 
> String var
> RAM:   [=====     ]  47.7% (used 39092 bytes from 81920 bytes)
> Flash: [====      ]  44.1% (used 461064 bytes from 1044464 bytes)
> 
> char[]
> RAM:   [=====     ]  47.7% (used 39104 bytes from 81920 bytes)
> Flash: [====      ]  44.1% (used 461052 bytes from 1044464 bytes)
>

MarcoCLA: I updated according to your suggestion.
This will decrease ram usage, however it will just dynamically allocate the value and put it on the heap and increase fragmentation. As it is a value with a known fixed length I personally think a char array fits better.

universam1: Not opposed to canonical char[] at all, but then lets reduce it to the minimum which is also the leanest of all

> RAM:   [=====     ]  47.7% (used 39104 bytes from 81920 bytes)
Flash: [====      ]  44.1% (used 460972 bytes from 1044464 bytes)
```suggestion
  char ssid[14];
  snprintf(ssid, sizeof(ssid), "iSpindel_%X", ESP.getChipId());
  wifiManager.startConfigPortal(ssid);
```

universam1:
I like the idea!
Would you mind adjusting the docs as well to reflect that new behaviour?

universam1 (approved):
excellent!

universam1:
Thanks a lot - LGTM

Issue #463, opened 2021-02-09
Possible bufferoverflow in validateInput


MarcoCLA:
The validateInput function uses the length of the string instead of the length of the chararray.
https://github.com/universam1/iSpindel/blob/499d358af8423906653e9b0dd1e2bc69fa47b7f7/pio/src/iSpindel.cpp#L299

The length of the chararray is TKIDSIZE
https://github.com/universam1/iSpindel/blob/499d358af8423906653e9b0dd1e2bc69fa47b7f7/pio/src/iSpindel.cpp#L70
The parameter is initialized as TKIDSIZE x 2
`  
WiFiManagerParameter custom_name("name", "iSpindel Name", htmlencode(my_name).c_str(),
                                   TKIDSIZE * 2);
`
So it will strncpy TKIDSIZE x 2 instead of the TKIDSIZE chararray in the validateInput function. This happens for multiple parameters.

[toCharArray ](https://github.com/esp8266/Arduino/blob/8fe80f163063470bf2a3188cd827f74260f42283/cores/esp8266/WString.h#L233) and [getBytes](https://github.com/esp8266/Arduino/blob/8fe80f163063470bf2a3188cd827f74260f42283/cores/esp8266/WString.cpp#L551) for reference.

universam1:
Thanks for the investigation, would you like to create a fix?

PR #464, opened 2021-02-09
Align lenghts to chararrays

MarcoCLA:
Hotfix for #463, if lengths are set correctly in WiFiManagerParameter they do not really have to be validated in validateInput


universam1 (approved):
good catch!

Issue #469, opened 2021-02-21
Possible division by Zero in tilt calculation


pppedrillo:
For bad/dodgy gyro/accel modules it is possible that all the readings (ax, ay and az) are all zeroes.
In that case calculateTilt() returns NaN.
**_az / (sqrt(_ax * _ax + _ay * _ay + _az * _az))**  --> division by 0.

The same possible issue with **void setup()** function, when USE_DMP defined - the same calculations copy-pasted there.

iSpindel.cpp
--------------------------------------
float calculateTilt()
{
  float _ax = ax;
  float _ay = ay;
  float _az = az;
  return acos(_az / (sqrt(_ax * _ax + _ay * _ay + _az * _az))) * 180.0 / M_PI;
}

void setup()
{

  Serial.begin(115200);

....
    ax = euler[0];
    ay = euler[2];
    az = euler[1];

    float _ax = ax;
    float _ay = ay;
    float _az = az;
    Tilt = acos(_az / (sqrt(_ax * _ax + _ay * _ay + _az * _az))) * 180.0 / M_PI;
  }
#endif
...
}

PR #470, opened 2021-02-21
Possible division by zero in tilt calculations fix for issue 469 

pppedrillo:
Fix for issue https://github.com/universam1/iSpindel/issues/469

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -855,10 +855,10 @@ void initAccel()
 
 float calculateTilt()
 {
-  float _ax = ax;
-  float _ay = ay;
-  float _az = az;
-  return acos(_az / (sqrt(_ax * _ax + _ay * _ay + _az * _az))) * 180.0 / M_PI;
+  if ( ax==0 && ay == 0 && az == 0)
+        return 0.f;
+  
+  return acos(az / (sqrt(ax * ax + ay * ay + az * az))) * 180.0 / M_PI;
 }
 
 bool testAccel()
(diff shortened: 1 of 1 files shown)


MarcoCLA:
The refactor of the code I agree with.
But why would you want to return 0 if no angle could be determined?

pppedrillo:
Sure we can return anything in this case - the value doesnt make sense anyway.
But "0" is easy to spot - eg user will rotate iSpindel and see tilt is still "0" (not a "NaN" like it is now). It was like that in the version up to 6.3 I believe, before this "improved" formula with came in - tilt was zero in case of hw failure.

ErikdBr:
> Sure we can return anything in this case - the value doesnt make sense anyway.
> But "0" is easy to spot - eg user will rotate iSpindel and see tilt is still "0" (not a "NaN" like it is now). It was like that in the version up to 6.3 I believe, before this "improved" formula with came in - tilt was zero in case of hw failure.

It was indeed zero up to 6.3-ish. Zero is better then NaN.

MarcoCLA:
In my opinion returning NaN (or something else which is not a valid value) is better in case of a hardware failure and can be easily detected at boot.
An angle of 0 is technically a valid angle and can occur. A acceleration of 0 on all three axis (which causes the error) not.
If the choice is made to return 0, will the temperature sensor code be changed to be in line with the accelgyro? Then that one should also just return 0 if no sensor is found (currently the iSpindel will just not continue booting).

It does not really matter which of the two implementations is chosen, but I think they should be the same, so the two choices:
- Stop booting if a problem with the temp/accel sensor is detected
- Just return 0 if the sensor is problematic.

pppedrillo:
> Then that one should also just return 0 if no sensor is found (currently the iSpindel will just not continue booting).

I completely agree and frankly couldn't understand from the very beginning why iSpindel is treating missing or not working temp sensor that specific way.

iSpindel can work without accel/gyro as a complicated thermometer.
iSpindel can work without temp probe as a hydrometer.
Both usecases make sense to me.

universam1 (approved):

PR #471, opened 2021-02-24
Sender: Ubidots handler

jotathebest:

### DIFF
--- a/pio/lib/Sender/Sender.cpp
+++ b/pio/lib/Sender/Sender.cpp
@@ -11,7 +11,7 @@
 #include <ThingSpeak.h>
 #include <BlynkSimpleEsp8266.h> //https://github.com/blynkkk/blynk-library
 
-#define UBISERVER "things.ubidots.com"
+#define UBISERVER "industrial.api.ubidots.com"
 #define BLYNKSERVER "blynk-cloud.com"
 #define CONNTIMEOUT 2000
 
@@ -381,7 +381,9 @@ bool SenderClass::sendUbidots(String token, String name)
         msg += name;
         msg += "?token=";
         msg += token;
-        msg += F(" HTTP/1.1\r\nHost: things.ubidots.com\r\nUser-Agent: ESP8266\r\nConnection: close\r\nContent-Type: application/json\r\nContent-Length: ");
+        msg += F(" HTTP/1.1\r\nHost: ");
+        msg += UBISERVER;
+        msg += "\r\nUser-Agent: ESP8266\r\nConnection: close\r\nContent-Type: application/json\r\nContent-Length: ");
         msg += measureJson(_doc);
         msg += "\r\n";
 


universam1 (approved):
Thank you

thegreatgunbantoad:
This doesn't compile.  There's a coding error in sender.cpp on line 386.  Accidental ')' at the end I think.

jotathebest:
@thegreatgunbantoad thanks, my bad. Fixed at https://github.com/universam1/iSpindel/pull/480

thegreatgunbantoad:
@jotathebest there is possibly another issue here.  It seems there are actually 2 servers for Ubidots:
```
static UbiServer UBI_INDUSTRIAL = "industrial.api.ubidots.com";
static UbiServer UBI_EDUCATIONAL = "things.ubidots.com";
```
Code from: https://github.com/ubidots/ubidots-esp8266/blob/master/src/UbiConstants.h

So all this does is switch over to the industrial server which is the paid for one.  The free accounts still use the things one I believe.

Looks like we need need an option in the AP so you can select your server.

Annoyingly I just can't get anything to connect to my Ubidots with either server.

ErikdBr:
Although Ubidots educational is deprecated, according to this article, Ubidots STEM will work with the current and earlier versions of the firmware 6.5.1. So just changing the URL to the industrial is wrong. Prior educational users have to signup for a STEM account.
https://help.ubidots.com/en/articles/3979278-connect-the-ispindel-low-cost-diy-hydrometer-with-ubidots

camsaway:
Perhaps rather make the server option visible ubidots instead of hidden. Default to the free version but will allow users to change to the old URL or paid for URL.

jotathebest:
@thegreatgunbantoad since the deprecation of Ubidots for education, both STEM and Industrial users must send data to `industrial.api.ubidots.com` . The old endpoint, `things.ubidots.com`, will still be available to allow users that have already deployed devices in the field to ingest data to Ubidots, but we officially advise using the industrial endpoint, which has higher availability and supports all the new features of our REST API v2.0

The code of the ESp8266, that you have just shared, will also be updated in the incoming weeks to delete any reference to old educational endpoints. Actually, the most updated library is the one developed for [Particle devices](https://github.com/ubidots/ubidots-particle), we hope to update the ESP8266 and ESP32 libraries soon too.

Let me know if this PR fixes the issue related to the missing parenthesis. If not, I will be happy to fork the entire repository again in order to update the issued patch properly.

@ErikdBr the article that you have shared is outdated, I will ask the team to update the endpoint to ingest data already written in the article. The current commit **will work** for both STEM and industrial users.

thegreatgunbantoad:
@jotathebest ahh very interesting.  My Ubidots still just refuses to play so I can't really comment on which one works.  It used to work until I switched to STEM (I think that was it, haven't used it for a long while due to the old fees being very high).  Now it just refuses to accept data it seems.  The setup isn't really that hard so I'm not sure what I could have messed up.

PR #480, opened 2021-03-19
Sender: Ubidots handler

jotathebest:


universam1 (changes requested):
There are merge conflicts, please resolve, thanks

PR #485, opened 2021-03-26
Now can change modes by tilting the sensor

thormj:
Now can change modes by tilting the sensor
 > 145 (on cap) = Deep Sleep (no RF)
 > 100 => Config
Others; no change.

Also changed the max sleep interval to 30 seconds so that it wakes up every 30 sec just
enough to see if there's a new tilt command; if not, it does the "next sleep step"...

### DIFF
--- a/pio/lib/Globals/Globals.h
+++ b/pio/lib/Globals/Globals.h
@@ -20,7 +20,7 @@
 extern Ticker flasher;
 
 // defines go here
-#define FIRMWAREVERSION "6.5.1"
+#define FIRMWAREVERSION "6.5.2"
 
 #define API_FHEM true
 #define API_UBIDOTS true
@@ -109,9 +109,12 @@ extern Ticker flasher;
 
 // sleep management
 #define RTCSLEEPADDR 5
-#define MAXSLEEPTIME 3600UL //TODO
+#define MAXSLEEPTIME 30UL //Check the accelerometer every 30 secs for a command   //3600UL //TODO
 #define EMERGENCYSLEEP (my_sleeptime * 3 < MAXSLEEPTIME ? MAXSLEEPTIME : my_sleeptime * 3)
 #define LOWBATT 3.3
+
+// tilt command
+#define MODERTCSLOT 20
 
 #define UNINIT 0
 
--- a/pio/lib/Globals/Globals.h
+++ b/pio/lib/Globals/Globals.h
@@ -108,13 +108,15 @@ extern Ticker flasher;
 #define RTCVALIDFLAG 0xCAFEBABE
 
 // sleep management
-#define RTCSLEEPADDR 5
-#define MAXSLEEPTIME 30UL //Check the accelerometer every 30 secs for a command   //3600UL //TODO
+#define RTCSLEEPADDR 5      //And +1 for the valid flag
+#define MAXSLEEPTIME 30UL   //Check the accelerometer every 30 secs for a command   //3600UL //TODO
 #define EMERGENCYSLEEP (my_sleeptime * 3 < MAXSLEEPTIME ? MAXSLEEPTIME : my_sleeptime * 3)
 #define LOWBATT 3.3
 
 // tilt command
-#define MODERTCSLOT 20
+#define RTCMODE 10
+enum modes {locknormal, normal, config, deepsleep};
+extern String tiltcommands[];
 
 #define UNINIT 0
 
--- a/pio/lib/WiFiManagerKT/WiFiManagerKT.cpp
+++ b/pio/lib/WiFiManagerKT/WiFiManagerKT.cpp
@@ -957,6 +957,9 @@ void WiFiManager::handleInfo()
   DEBUG_WM(F("Sent info page"));
 }
 
+String tiltcommands[]={"Normal","Config","DeepSleep"};
+extern volatile String Pronk;
+
 /** Handle the info page */
 void WiFiManager::handleiSpindel()
 {
@@ -964,6 +967,8 @@ void WiFiManager::handleiSpindel()
 
   // we reset the timeout
   _configPortalStart = millis();
+  int tiltcommand;
+  ESP.rtcUserMemoryRead(MODERTCSLOT, (uint32_t *)&tiltcommand, 1);
 
   String page = FPSTR(HTTP_HEADER);
   page.replace("{v}", "Info");
@@ -987,6 +992,10 @@ void WiFiManager::handleiSpindel()
   page += F("V</td></tr>");
   page += F("<tr><td>Gravity:</td><td>");
   page += String(Gravity, 3);
+  page += F("</td></tr>");
+  page += F("<tr><td>TiltCommand:</td><td>");
+  page += tiltcommands[tiltcommand];
+  page += F("</td></tr>");
   page += F("</td></tr>");
   page += F("</table></h2>");
   page += F("<hr><dl>");
@@ -1028,7 +1037,7 @@ void WiFiManager::handleMnt()
   page += F("<dd>Date: ");
   page += __DATE__ " " __TIME__;
   page += F("</dd></dl><br>");
-  page += F("<form method='POST' action='/update' enctype='multipart/form-data'><input type='file' name='update'><br><input type='submit' class=\"btn\" value='update'></form>");
+  page += F("<form method='POST' action='/update' enctype='multipart/form-data'><input type='file' name='update' accept='.bin'><br><input type='submit' class=\"btn\" value='update'></form>");
   page += F("<hr><h2>Factory Reset</h2><br>All settings will be removed");
   page += F("<br><form action=\"/reset\" method=\"get\"><button class=\"btn\">factory reset</button></form><br/>");
   page += FPSTR(HTTP_END);
--- a/pio/lib/WiFiManagerKT/WiFiManagerKT.cpp
+++ b/pio/lib/WiFiManagerKT/WiFiManagerKT.cpp
@@ -957,9 +957,6 @@ void WiFiManager::handleInfo()
   DEBUG_WM(F("Sent info page"));
 }
 
-String tiltcommands[]={"Normal","Config","DeepSleep"};
-extern volatile String Pronk;
-
 /** Handle the info page */
 void WiFiManager::handleiSpindel()
 {
@@ -968,7 +965,7 @@ void WiFiManager::handleiSpindel()
   // we reset the timeout
   _configPortalStart = millis();
   int tiltcommand;
-  ESP.rtcUserMemoryRead(MODERTCSLOT, (uint32_t *)&tiltcommand, 1);
+  ESP.rtcUserMemoryRead(RTCMODE, (uint32_t *)&tiltcommand, 1);
 
   String page = FPSTR(HTTP_HEADER);
   page.replace("{v}", "Info");
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -951,7 +951,11 @@ void initAccel()
 float calculateTilt()
 {
   if (ax == 0 && ay == 0 && az == 0)
-    return 0.f;modes curmode;
+    return 0.f;
+
+  float angle = acos(az / (sqrt(ax * ax + ay * ay + az * az))) * 180.0 / M_PI;
+
+  modes curmode;
   ESP.rtcUserMemoryRead(RTCMODE, (uint32_t *)&curmode, sizeof curmode);
 
   if(curmode != locknormal)
@@ -996,7 +1000,8 @@ float calculateTilt()
     }
   }
 
-  return angle;}
+  return angle;
+}
 
 bool testAccel()
 {
@@ -1377,7 +1382,6 @@ void setup()
     // decide whether we want configuration mode or normal mode
     if (shouldStartConfig(validConf))
     {
-does this make sense?
       {
         // test if ssid exists
         if (WiFi.SSID() == "" &&
@@ -1444,8 +1448,6 @@ does this make sense?
       CONSOLE(euler[1] * 180 / M_PI);
       CONSOLE("\t");
       CONSOLELN(euler[2] * 180 / M_PI);
-
-Does This Make Sense?
 
       float _ax = ax;
       float _ay = ay;
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -92,7 +92,7 @@ int8_t my_OWpin = -1;
 int usequicksleep;
 
 // Tilt-commanded modes
-#define COMMANDANGLE 100.0   //"tilted above"; should never ever happen when being used.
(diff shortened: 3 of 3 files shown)


ErikdBr:
What is the usecase for this? Cause I am sorry but I don't see it. I don't want my iSpindel magically change modes just by tilting it.

thegreatgunbantoad:
I do sort of like the idea of upending it popping it into config. (Sorry couldn't be bothered to scour all the code to see the specifics of how you've implemented it) Presumably you'd have to leave it upended for at least 1 sleep period though, so I'd get bored and pop the cap.  There is a hardware solution that I have done that works well which is to fit a reed switch on the rest pins and just wave a magnet at it.  I even have one with a wireless charge pad in it so you never need to open the container (However the wireless charge takes ages so I do open it to charge it anyway).

Since you'd have to take the iSpindel out of the fermenter to upend it though I don't see why you wouldn't just pop the cap.
Also I very rarely use the config, I use it as more of a set and forget device really.  Like @ErikdBr  I don't really see the use case. Interesting idea, I'm a bit on the fence at the moment.

ErikdBr:
I use iSpindels as wifi thermometers standing on the cap

thegreatgunbantoad:
> I use iSpindels as wifi thermometers standing on the cap

That's probably a bit niche.

MarcoCLA:
https://xkcd.com/1172/ xD

It's an innovative idea, but I don't really see the use case either. I configured the iSpindel once and after that I didn't really have to touch the config.

pio/lib/WiFiManagerKT/WiFiManagerKT.cpp
```diff
@@ -1028,7 +1037,7 @@ void WiFiManager::handleMnt()
   page += F("<dd>Date: ");
   page += __DATE__ " " __TIME__;
   page += F("</dd></dl><br>");
-  page += F("<form method='POST' action='/update' enctype='multipart/form-data'><input type='file' name='update'><br><input type='submit' class=\"btn\" value='update'></form>");
+  page += F("<form method='POST' action='/update' enctype='multipart/form-data'><input type='file' name='update' accept='.bin'><br><input type='submit' class=\"btn\" value='update'></form>");
```
universam1, line 1040: nice little improvements but somewhat unrelated?

universam1 (changes requested):
Thank your for submitting this idea!
I'm unsure if it will help the user experience, in my case it won't - but I'm open for it if it is guarded behind an config option. 
If you'd like to come up with an implementation that contains this feature that can be enabled (disabled by default) I'm happy to accept.

Please note that your branch is outdated.

ErikdBr:
> > I use iSpindels as wifi thermometers standing on the cap
> 
> That's probably a bit niche.

You probably mean 'a bit snobbish' lol.

pio/lib/WiFiManagerKT/WiFiManagerKT.cpp
thormj: Yep.  very unrelated.  I just kept trying to upload the ELF instead of the BIN file... so I change it so it would default to the correct extension.

thormj:
Hrm... GitHub's threading leaves a lot to be desired.
I did this because I attached an inductive charger to the port (https://amzn.to/39pLT1w) and wanted to "never open it again."

So, "tits up" it goes to sleep (it's actually still reading the accelerometer and the battery every 30 sec, but doesn't turn on the WiFi) -- I just store them on their cap when not in use and they blink every 30 sec (now to 3d print a vertical Qi charging base).

If I want to change it (change the interval or do some more development [eg, this], I need to be able to get into "Config" mode without having to short some pins... so, if you put it at an "impossible for normal use" angle, it goes into config mode.  You can let the watchdog knock it out or you can go to the webpage and say "save configuration" to exit (just like if you went in via double-reset).

--
Small plans ahead:
1. Attach vibration motor and have a config for that.
2. Change the accelerometer code so it only wakes up the accel part of the device (150uA vs 4 mA)

So... it looks like I should wrap this in a config wrapper; maybe add a "update wifi in sleep every XX seconds"...

ErikdBr:
> I did this because I attached an inductive charger to the port (https://amzn.to/39pLT1w) and wanted to "never open it again."

I tried to implement a wireless charger but because you have to bend the receiver into the tube it got so hot it almost caught fire.

To "never open it again" will result in that your 2nd brew with the same iSpindel will get infected because you'll never be able to sanitize it properly. Yeast and other residues will accumulate between the thread of the cap and the thread of the tube. I always sanitize the cap seperate from the tube.

thormj:
Fixed, updated, merged, fixed and pushed (I hope)...

Now there are 2 settings on the config page-
Enable (it's a number but anything besides 0 enables "Tilt Commands")
Deep Sleep Update Interval -- if it's on the base it will update at _this interval_ instead (so if you want your iSpindels to log their battery state to ubidots once a day while in storage, you can do that.

thormj:
So... I'm not sure what lint is complaining about, but I'll be glad to fix it.

thormj (commented):
Added the config guard and added a "on-cap-wakeup time" that can be different from the "normal wakeup time."

MarcoCLA:
This are still quite a lot of changes needed for such a little usecase. 
If this is even included in master, I would prefer to make this optional in the binary using a compiler flag, as a lot of people will not need or want this.

ErikdBr:
> as a lot of people will not need or want this.

If this is optional in the config page by checking a box or swiping a switch and the rest of the options related to this is greyed out or even better hidden, fine.
I think most of the brewers with one iSpindel will flip the on/off switch to turn it on and throw it in their brew and after the brew turn it off and charge it, and so don't need this.

ErikdBr:
> Deep Sleep Update Interval -- if it's on the base it will update at _this interval_ instead (so if you want your iSpindels to log their battery state to ubidots once a day while in storage, you can do that.

So if it is on the cap and this is left blank it will still report everything to whatever service used at the normal interval?

thormj:
> So if it is on the cap and this is left blank it will still report everything to whatever service used at the normal interval?
The default is once an hour.  If blank, I think it it would be read as zero and I'm not sure how that would react for either the "normal" or "sleeping" interval.

ErikdBr:
> **if** it's enabled with a non-zero _Tilt Command Enabled_

So this feature is optional and thus can be turned on or off?

thormj:
> This are still quite a lot of changes needed for such a little usecase.

"linecount-wise," it seems like a lot, but that's because the original code (and even this a bit) is a bit haphazard as to "where I invoke goodNight" or "do I need to make sure the radio is on" or other stuff.

I tried to move, recompartmentalize some of the blocks (like "EnsureRadioOn") and just add a bit to the "figure out tilt" and a couple of "responsiveness bits" near/around the goodNight area.

> If this is even included in master, I would prefer to make this optional in the binary using a compiler flag, as a lot of people will not need or want this.

If they don't enable it (with a 1 in the configuration page), it doesn't do anything new (I think I cleaned up a bit of the "Wake up radio" and "goodNight" bits).  But... I did this because I want to be able to take the device out, clean it, place on charger (vertically so it goes to sleep [a la commercial] -- my Qi charger doesn't seem to mind bending, but I also have one I tried that fits into the cap)... sanitize and throw back into the bucket.  When we get a new feature (maybe the hierarchical menus or a "interval schedule" so that early fermentation is often, but late is not so much) or something, you can tilt it up wrongways and load new firmware into it (or load a new polynomial or change update rates since you want to do mead now instead of beer and you need 6mo battery life -- I haven't really test long runs yet) without reed relays or a shorting screwdriver or fumbling for the reset switch).

On the SW side (this implementation is a bit simple, but...) we could use the accelerometer for other things ("tilt wrong way battery up = do x, wrong way down do y, right way battery up [also unpossible], do z, etc.)  unfortunately this accelerometer doesn't have a "wake-on-shake" feature so that's why I changed the max-interval to 30 sec -- to reduce the time between a command and its action).

Sounds better than the commercial offering where I have to take it apart to change the batteries every 3-4 brews.
Now on to adding vibration; my last wine had so many bubbles entrained in it it fubarred the commercial device until I racked it.

Then on to making the battery life... great (debug levels, only wake up the accelerometer, etc -- keep the "woke time" as small as possible, and probably that "interval schedule" [if SG is >x, update every hr, >y update every 4, near end, every 8]).

thormj:
Well, I enabled the clang-format... but it didn't make any changes when I told it to Format Document... :/

MarcoCLA:
> is a bit haphazard

> I tried to move, recompartmentalize some of the blocks (like "EnsureRadioOn") and just add a bit to the "figure out tilt" and a couple of "responsiveness bits" near/around the goodNight area.

 > I still don't quite like the RTC stuff being scattered everywhere but... that's a next step.

I totally agree, I already started cleaning stuff in #491 and #492. Adding new features is not bad, but in the current state of the project hard to do and to compartmentalize this functionality without changing code all over the place.
I think the project as a whole needs some intensive refactoring as currently state it is a bit messy. Your steps in that regard are great! But still a lot can be done. Maybe we can both work on this together.

Tip try something like:
```clang-format ./pio/src/*.cpp --style=file -i```

**Comment by ErikdBr** [comment]
> If they don't enable it (with a 1 in the configuration page), it doesn't do anything new

But does it still do what it allways has been doing, sending the data no matter what angle?

**Comment by thormj** [comment]
> But does it still do what it allways has been doing, sending the data no matter what angle?

Yeo,,,  I think it should be slightly more reliable at figuring out "is the radio on" vs not (I set the radio on/off explicitly in goodNight), but if the Tiltcmden is set to 0, the RTC is set to "HardNormal"... and most of the code is bypassed.

The only change in behavior is that it will wake up (no radio) and go back to sleep every 30 seconds and read the accelerometer and battery (since I set MAXSLEEPTIME to 30 instead of LargeNumber),,, but the "Accept command" in the "Read accelerometer" is bypassed if the mode is HardNormal (it will print more stuff to the serial port) -- it has to be in the "Read accelerometer" part because in config mode, that's tied to a blinktimer instead of being a mainline loopy thing.

**Comment by thormj** [comment]
> #492. Adding new features is not bad, but in the current state of the project hard to do and to compartmentalize this functionality without changing code all over the place.

Aye, that's always the "gotcha" with embedded thingies; they need access to all kinds of different parts I kept #492 in here because it's a great example -- the DRD "just" needs access to the RTC... but so does everybody else that wants a NV-but-not-flash store... 

So I was thinking of making some kind of "wrapper" that handled the RTC and made it easy (in the Globals file) to add/move) the RTC registers and see what's used and what's not (hint... there's a plus one in there too)...
--
Thanks for the clang-format tip.  Now clang-formatted.

File: pio/lib/Globals/Globals.h [hunk 2]
```diff
@@ -110,11 +110,16 @@ extern Ticker flasher;
 #define RTCVALIDFLAG 0xCAFEBABE
 
 // sleep management
-#define RTCSLEEPADDR 5
-#define MAXSLEEPTIME 3600UL //TODO
+#define RTCSLEEPADDR 5      //And +1 for the valid flag
+#define MAXSLEEPTIME 30UL   //Check the accelerometer every 30 secs for a command   //3600UL //TODO
```
universam1, line 114: this point makes it already clear that this idea reduces the battery lifetime by a factor of ~10x!
No matter the optimization that is a significant issue that make this idea hard to happen

universam1 (changes requested):

universam1:
There is one practical point often forgotten with the "I never want to open the iSpindel any more" idea which comes up every other month.
Simply the point that old yeast/bacteria will stay in the cap and will ruin your next batch. It is from my experience practically not possible to clean up the iSpindel "Petling" so well that nothing resides in the cap.
Just for hygiene reasons this does not work from what I learned

ErikdBr:
> There is one practical point often forgotten with the "I never want to open the iSpindel any more" idea which comes up every other month.
> Simply the point that old yeast/bacteria will stay in the cap and will ruin your next batch. It is from my experience practically not possible to clean up the iSpindel "Petling" thread so well that nothing resides in the cap.
> Just for hygiene reasons this does not work from what I learned

I totally agree as I stated before: https://github.com/universam1/iSpindel/pull/485#issuecomment-809566874

thormj:
> Simply the point that old yeast/bacteria will stay in the cap and will ruin your next batch. It is from my experience practically not possible to clean up the iSpindel "Petling" thread so well that nothing resides in the cap.

And I thank you for that knowledge... so instead of trying it first (and screwing up a brew), I'll design a new cap for the Petling so that it seals against the "grip ring", closing off the threads.  And I'll probably dip the top in silicone just to be sure  Or maybe get a food safe heat shrink cap and ? -- but "food safe" and "heat shrink" and "waterproof" sound contradictory).

This patch still lets me "keep upright to go to sleep" (a la commercial, but the commercial battery isn't rechargeable) and gives me a way to get into "config mode" without having to take it apart and short it... If there are other things that people want (dunno what though...), we can extend the calculateTilt to return negative numbers for "battery up" [which won't happen in real life either) "normal" or "config" so we could send "3" commands based on angle (maybe add some for "hey, I'm being shaken") but,,, I don't know what else we'd want besides "sleep" (save battery) and "go to config; I have new poly or I want to change update or I want to update firmware").
 
 I'm setting up an expieriment to see what the power usage has changed to::
 Fully charge one and get the battery voltage, set the deep sleep update to twice a day (just to track things); set it on end.
 Take 5 days of readings; calculate the point at which the battery voltage = 3.3V; and that's the max lifetime with this patch

I also want to rig it up to my SiLabs AEM system and get real power draws so I can really track the effect of power saving techniques....

So rework is "in progress."  You can change the number for MaxSleepInterval back, but then there would be up to an hour from changing the orientation of the sensor 'till it started doing what you asked (for "normal use," that's probably okay; but I want to flip it around and go into config mode to upload new firmware).  I'm looking into how to wire the MPU to the reset so we could change it back and then "shake to get its attention."

pppedrillo:
>  so we could change it back and then "shake to get its attention."

This might not work either - during active fermentation phase iSpindel have a really proper shaking and smashes against the fermenter's  walls.

Sidenote about decreasing battery time more than x100 times : it means iSpindel battery **will last only 1 day** instead of 3 months . Does it make any sense?

thormj:
> This might not work either - during active fermentation phase iSpindel have a really proper shaking and smashes against the fermenter's walls.

Ah... that's good to know, so I might need to put some logic in it (I was going to set the "shake" really high, but an accell from hitting a wall is actually wayy higher than "shakeweight reenactment."

> Sidenote about decreasing battery time more than x100 times : it means iSpindel battery **will last only 1 day** instead of 3 months . Does it make any sense?

Hasn't been my experience... my Red and my White have been going for about 2 weeks since I added this.  Remember, the "wake to read accelerometer" is a "wake without radio" and is *much* shorter (2 accelsamples and no DS1820 action) than a "Working Shift."

I really need to get this hooked to my AEM test rig so I can share the data for all the bits and pieces.

PR #488, opened 2021-03-27
Add code lint

MarcoCLA:
No clue which actual format was used, but this is the closest I got.
@universam1 what format is used?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,7 @@
+---
+BasedOnStyle: Microsoft
+IndentWidth: '2'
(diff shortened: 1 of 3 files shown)


universam1 (approved):
Excellent

universam1:
Thanks @MarcoCLA really like this linting improvement!

universam1 (changes requested):
Would you mind fixing the git conflict then happy to merge

universam1 (approved):
:+1:

PR #491, opened 2021-03-30
Cleanup dallas

MarcoCLA:
Remove a lot of redundant code from DS18B20 init and request temp code. Nearly all of this code is already in the dallas lib. 
Same OneWire autodetection and level of (usefull) debug information is kept.

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -839,31 +839,54 @@ void sleepManager()
 
 void initDS18B20()
 {
-  for (uint8_t p = 0; p < sizeof(OW_PINS); p++)
+  //Try using old pin if set
+  if (my_OWpin != -1)
   {
-    const byte pin = OW_PINS[p];
-    CONSOLE(F("TempInit: scanning for OW device on pin: "));
-    CONSOLELN(pin);
-    oneWire = new OneWire(pin);
+    CONSOLE(F("TempInit: scanning for saved OW device on pin: "));
+    CONSOLELN(my_OWpin);
+    oneWire = new OneWire(my_OWpin);
     DS18B20 = DallasTemperature(oneWire);
-    if (DS18B20.getAddress(tempDeviceAddress, 0))
+    if (!DS18B20.getAddress(tempDeviceAddress, 0))
     {
-      CONSOLELN(F("Tempinit: OneWire sensor found!"));
-
-      if (DS18B20.isConnected(tempDeviceAddress))
+      CONSOLE(F("TempInit: not found on: "));
+      CONSOLELN(my_OWpin);
+      CONSOLELN(F(" -> Reset my_OWpin "));
+      my_OWpin = -1;
+    }
+  }
+  //Find pin
+  if (my_OWpin == -1)
+  {
+    for (uint8_t p = 0; p < sizeof(OW_PINS); p++)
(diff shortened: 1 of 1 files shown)


pio/src/iSpindel.cpp
```diff
@@ -786,60 +786,38 @@ void sleepManager()
...
-  {
-    DS18B20.requestTemperatures();
-    DSreqTime = millis();
-  }
-}
-
 void initDS18B20()
 {
-  if (my_OWpin == -1)
+  for (uint8_t p = 0; p < sizeof(OW_PINS); p++)
```
universam1, line 791: this would perform a scanning of pins on every wake up which is a time/battery waste and potential runtime instability.
I don't see a reason why this is better than the initial action during setup and only happening at that supervised occasion.

universam1 (changes requested):

pio/src/iSpindel.cpp
MarcoCLA: Fixed

MarcoCLA:
Go time to review this?

pio/src/iSpindel.cpp
```diff
@@ -786,60 +786,61 @@ void sleepManager()
...
-      delete oneWire;
-      oneWire = new OneWire(my_OWpin);
+      const byte pin = OW_PINS[p];
+      CONSOLE(F("TempInit: scanning for OW device on pin: "));
+      CONSOLELN(pin);
+      oneWire = new OneWire(pin);
       DS18B20 = DallasTemperature(oneWire);
-      DS18B20.begin();
-      DS18B20.getAddress(tempDeviceAddress, 0);
+      if (DS18B20.getAddress(tempDeviceAddress, 0))
```
kmalisiewicz, line 816: This **if** statement can be much more readable when you remove **elses**:

```
  if (!DS18B20.getAddress(tempDeviceAddress, 0))
  {
    CONSOLELN(F("Tempinit: No OneWire sensor found"));
    break;
  }

  CONSOLELN(F("Tempinit: OneWire sensor found!"));

  if (!DS18B20.isConnected(tempDeviceAddress))
  {
    CONSOLELN(F("Tempinit: sensor NOT connected"));
    break;
  }
  
  CONSOLELN(F("Tempinit: sensor connected"));

  if (!DS18B20.validFamily(tempDeviceAddress))
  {
    CONSOLELN(F("Tempinit: temp sensor family INVALID"));
    break;
  }
  
  CONSOLELN(F("Tempinit: temp sensor family VALID"));
  my_OWpin = pin;
  break;
```

pio/src/iSpindel.cpp
```diff
@@ -837,60 +837,62 @@ void sleepManager()
...
+      CONSOLELN(pin);
+      oneWire = new OneWire(pin);
       DS18B20 = DallasTemperature(oneWire);
-      DS18B20.begin();
-      DS18B20.getAddress(tempDeviceAddress, 0);
+
+      if (!DS18B20.getAddress(tempDeviceAddress, 0))
+      {
+        CONSOLELN(F("Tempinit: No OneWire sensor found"));
+        break;
```
kmalisiewicz, line 871: I was wrong with **break** here. Here should be continue and after `my_OWpin = pin;` should be break.

PR #492, opened 2021-03-30
Move DRD to lib

MarcoCLA:
More cleanup


pio/src/iSpindel.cpp
```diff
@@ -1170,7 +1173,6 @@ void setup()
     // DIRTY hack to keep track of WAKE_RF_DEFAULT --> find a way to read WAKE_RF_*
     if (tmp != RTCVALIDFLAG)
     {
-      drd.setRecentlyResetFlag();
```
universam1, line 1173: please explain this change

universam1, line 1173: please explain this change

universam1 (changes requested):

pio/src/iSpindel.cpp
MarcoCLA: That function is private and should not be used
https://github.com/datacute/DoubleResetDetector/blob/994c20de15e650cc488b8e519b839c81b3528c13/src/DoubleResetDetector.h#L36
Same behaviour is achieved with drd.detectDoubleReset() and drd.stop()

PR #493, opened 2021-03-30
Hide AP password in the config page and on console

wvolz:
These commits remove display of WiFi passwords from the serial console and config pages.

The wifimanagerKT config page was missing the type="password" setting on the wifi password input field. This hides display of the wifi password. Also, if you configure the config access point to use a password, this password was shown on the console. This has been commented out but left in case someone needs it for debugging.

Note: I have not tested either of these on a board yet. Will update once I've confirmed they are working ok.

### DIFF
--- a/pio/lib/WiFiManagerKT/WiFiManagerKT.cpp
+++ b/pio/lib/WiFiManagerKT/WiFiManagerKT.cpp
@@ -692,8 +692,7 @@ void WiFiManager::handleWifi()
   page += FPSTR(HTTP_FORM_START1);
   page += _ssid;
   page += FPSTR(HTTP_FORM_START2);
-  // don't send existing password to web
-  //page += _pass;
+  page += _pass;
   page += FPSTR(HTTP_FORM_START3);
 
   char parLength[2];
--- a/pio/lib/WiFiManagerKT/WiFiManagerKT.cpp
+++ b/pio/lib/WiFiManagerKT/WiFiManagerKT.cpp
@@ -150,7 +150,8 @@ void WiFiManager::setupConfigPortal()
       DEBUG_WM(F("Invalid AccessPoint password. Ignoring"));
       _apPassword = NULL;
     }
-    DEBUG_WM(_apPassword);
+    //Don't show ap password on the console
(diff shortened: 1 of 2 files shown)


MarcoCLA:
Similar functionality is already available in [upstream WiFiManager ](https://github.com/tzapu/WiFiManager/blob/8bfc228695e8998bc3ff1ca0494ca0601bf5034f/WiFiManager.cpp#L1239).

pio/lib/WiFiManagerKT/WiFiManagerKT.cpp
```diff
@@ -691,7 +692,8 @@ void WiFiManager::handleWifi()
   page += FPSTR(HTTP_FORM_START1);
   page += _ssid;
   page += FPSTR(HTTP_FORM_START2);
-  page += _pass;
+  // don't send existing password to web
+  //page += _pass;
```
universam1, line 696: there was a long discussion on this and the desire was to be able to see the password - so lets populate that field. However I am ok to mask that field with asterisks

universam1 (changes requested):

pio/lib/WiFiManagerKT/WiFiManagerKT.cpp
wvolz: @universam1 Got it. Before I submit that, would it make sense to merge the upstream WifiManager code instead? As @MarcoCLA mentioned that has the fix for displaying the password as asterisks. I'm not sure about specific changes but it does look to be more actively maintained than the current library.

universam1: There are a couple of issues that I have solved manually, regarding the responsive UI and interactive elements. Would need to see if that is possible nowadays with upstream first, but that would be a longer goal

wvolz: Ok. I've re-enabled sending of the PW to the webUI but it's hidden with asterisks. I'll open another PR for the merge of upstream WifiManager after I have some time to test that.

PR #494, opened 2021-04-04
This makes the AP Name the "Name" given to it (if no name exists, it uses the chip id like the master branch).

thormj:
Simple, but I couldn't remember which was which based on the ChipID... so if you named it "Red", it becomes "iSpindel_Red"


thormj:
Ack... let's try that again rebasing to master instead of halfway up tiltcommands.  Sometimes Git gets the best of me....

PR #495, opened 2021-04-04
Use my_name in the AccessPoint SSID

thormj:
I couldn't keep which was which with the ChipID, so now it is "iSpindel_Red" (up to 32 chars total, according to WiFi standards).
Better fix for issue #460 .


universam1 (approved):
👍

universam1:
good idea, thank you!

MarcoCLA:
Default value is now "iSpindel_iSpindel000"?

PR #497, opened 2021-04-05
Added a "FAQ for Improvements" file so noobs can see the tried paths ……that never made it

thormj:
A place for things that are commonly seen as decent ideas but have serious issues in them (eg, in retrospect, the PETling threads are a haven for bacteria, scum, and villainy).

I figure at if things get worked out, they can be stuck in here (eg I'm testing the https://amzn.to/3wuqU7s Qi charger and have noticed that it is *very* picky about placement and if it's in the wrong place it can overheat)... so I'm not confident in it yet, but when I am I'll have a link to the charger, and also have a patch that has instructions to make a mount that forces it to be in the "right spot"

Or not... but at least there's a warning to those who want to attempt the path from those who have taken it before...


FAQ-Improvements.md
```diff
@@ -0,0 +1,21 @@
...
+		charging pad.
+    There are inductive chargers that fit inside the cap (eg: 
+    https://amzn.to/3wslaLO , but they haven't been tried yet).
+
+  b)The seal for the PETling is at the top/inside; yeast and bacteria 
+    will get in the "threaded area" and will be impossible to wash out.
+    Right now, it's recommended to take the cap off and clean/sanitize
+    the cap and the PETling separately.
+
+2. Doesn't krausen/CO2 affect the readings?
```
pppedrillo, line 17: You can add a dryhopping as well.

thormj: Looks good to me.  I dont' know about dryhopping...

FAQ-Improvements.md
```diff
@@ -0,0 +1,27 @@
+This is a list of "realworld" notes that have been encountered... so people
+can look at "cool new feature" and find what's known/been done before.
+
+1. "Let's seal it and use inductive charging!"
+  a)The Qi chargers are larger than the PETling so they have to be bent.
+		some Qi chargers get very unhappy about that (overheating/fire),
+		while others work but are *very* picky about placement on the Qi
+		charging pad.
+    There are inductive chargers that fit inside the cap (eg:
```
universam1, line 9: it has been tested and it technically does work kind of, but it is anything but clean rather hacky. So that statement is not totally correct

universam1 (changes requested):
good approach - lets move these informations in the https://github.com/universam1/iSpindel/blob/master/docs/FAQ-en.md file to one place

thormj:
> good approach - lets move these informations in the https://github.com/universam1/iSpindel/blob/master/docs/FAQ-en.md file to one place

Done!  That's a good place for it but I didn't see it at first so I put a note about the FAQ in the Readme.md file.

universam1 (approved):
👍

universam1:
thank you for submitting

PR #501, opened 2021-05-09
Specify Python version

FernandoGarcia:
Hi!

Specifying the Python version will prevent compiling error where `python` is not linked to `python3`.

```
File "git_rev.py", line 5
    print(f'\'-D FIRMWAREVERSION="{version}"\'')
                                              ^
SyntaxError: invalid syntax
OSError: 'python git_rev.py' exited 1:
  File "/home/fernandogarcia/.platformio/penv/lib/python3.7/site-packages/platformio/builder/main.py", line 177:
    env.SConscript("$BUILD_SCRIPT") 
```

Best regards.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -37,7 +37,7 @@ board = d1_mini
 framework = arduino
 monitor_speed = 115200
 upload_speed = 921600
-build_flags = !python git_rev.py
+build_flags = !python3 git_rev.py
 lib_deps =
   ArduinoJson@>5
   ESP Async WebServer
--- a/platformio.ini
+++ b/platformio.ini
@@ -19,7 +19,7 @@ board = d1_mini
 framework = arduino
 monitor_speed = 115200
 upload_speed = 115200
-build_flags = !python git_rev.py
+build_flags = !python3 git_rev.py
 lib_deps =
   ArduinoJson@>5
   ESP Async WebServer


universam1 (approved):
👍

universam1:
thanks!

thegreatgunbantoad:
I've found the inclusion of the python build flags just breaks the build process.  I must be missing something.  It's claiming it can't find python I have it installed on the PC and the VS code extension.

lbussy:
Unless you have a real edge case need, you should be using a shebang like `#!/usr/bin/python` or (preferably) `#!/usr/bin/env python`.  Seems like you are making a problem where one should not be.  You are not doing scientific analysis, you're just grabbing some info.

thegreatgunbantoad:
It does build just fine if you just comment out the `build_flags = !python3 git_rev.py`

lbussy:
I didn't look at that part.  The point is the same, this is a solution in search of a problem. :)

thegreatgunbantoad:
Yeah I tend to agree.  Especially since it breaks what was working fine.

pppedrillo:
This change has to be reverted because it breaks build. I have 3.9.5 installed and 
build_flags = !python git_rev.py
works just fine, while "python3" fires 
"Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.
OSError: 'python3 git_rev.py' exited 9009:" @universam1

pppedrillo:
As a workaround, I had to duplicate python.exe and rename in into python3.exe.
Oh boy....

Issue #405, opened 2020-07-18
Https + Authentication header


alejandroaap:
Would it be possible to enable https requests and send the token in the authentication header and not in the body like right now?

Issue #352, opened 2020-02-26
HTTPS support


Nr18:
I was trying to implement an API for the iSpindel but I noticed it does not have HTTPS support?

Are there any plans to do so?

Colin66:
Hi, I asked similar a while back and the answer was that the wemos doesn't have the power to handle the https authentication, so use of an intermediate service is necessary.  
I used IFTTT as an intermediary to forward data to google services which require HTTPS.

PR #503, opened 2021-05-24
added support for https POST

schuhmc:
Added a new sender to send POST data via HTTPS without  checking the certificate. This way it is possible to send data via POST to a https server. It is insecure, but the data sent here is typically not sensitive.

#405 #352


universam1 (approved):
reasonable approach

PR #504, opened 2021-05-25
TCP Data Fixes

rtlindne:
The response from the TCP server was not being processed, when monitoring the response was in ASCII byte values, by casting the _client.Read to a char the response because text and was processed correctly. 

When sending TCP the RSSI was missing from JSON in the latest version, tracked that back to the size of the StaticJsonDocument _doc was reduced to 265 which is not big enough for the TCP data being sent


universam1 (approved):
👍🏻

universam1:
Thank you for that contribution

Issue #500, opened 2021-05-09
Compiling error from git_rev.py


FernandoGarcia:
## I'm submitting a ...
  - [x ] bug report

## Did you follow the general troubleshooting steps first:

- [ x] Searched for [similar reports already](https://github.com/universam1/iSpindel/issues) ?

## Report
Hi!

The script `git_rev.py` will cause compiling error when the code is downloaded as `.zip` instead `git clone`.

The compiling error:

```
fatal: not a git repository (or any of the parent directories): .git
Traceback (most recent call last):
  File "git_rev.py", line 4, in <module>
    version = subprocess.check_output(tagcmd, shell=True).decode().strip()
  File "/usr/lib/python3.7/subprocess.py", line 411, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.7/subprocess.py", line 512, in run
    output=stdout, stderr=stderr)
```

The script searchs for `git flags` but files from `.zip` are not a `git` repository.

Please check [this ](https://community.platformio.org/t/pio-looking-for-py-in-a-cpp-project/21197)thread for more details.

Best regards.

pppedrillo:
Related to https://github.com/universam1/iSpindel/pull/486 @MarcoCLA

PR #510, opened 2021-06-10
PlatformIO build configurations tweak (resolves  universam1/iSpindel#500)

pppedrillo:
Changes to platforio.ini to address https://github.com/universam1/iSpindel/issues/500 issue

1. Added new build configuration  d1_mini_standalone to be able to build from zip without having git installed
2. d1_mini renamed to d1_mini_git
3. d1_mini_stage renamed to d1_mini_git_stage

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,10 +49,10 @@ jobs:
     - name: Install library dependencies
       run: pio lib -g install 1
     - name: Run PlatformIO
-      run: pio run -e d1_mini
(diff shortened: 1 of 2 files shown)


universam1 (approved):
I see now reason why not to use the git version, but since it's transparent I have no objection to support that

pppedrillo:
Thanks for merging! Why do not use git version - It will give ability to people, who don't have a git installed on their computers, download sources as ZIP and build them. Otherwise they would need to install git as well or their build will fail. @universam1

PR #511, opened 2021-06-11
Fix preprocessor directives to be able to turn APIs on and off if needs

pppedrillo:
#ifdef s changed to #if
It gives ability to turn certain API(s) on and off in Global.h 

`#define API_UBIDOTS true`
or
`#define API_UBIDOTS false`
etc


universam1 (approved):
good idea, thanks

Issue #507, opened 2021-06-09
No stable WIFI connection w/ FW 7.0.0 & 6.6.0


peterschrott:
## I'm submitting a ...
  - [x] bug report
  - [ ] feature request
  - [X] support request => Please note that this repositories issues are intended for feature requests and bug reports.

## Did you follow the general troubleshooting steps first:

- [X] Follow through the [FAQ](https://github.com/universam1/iSpindel/blob/master/docs/FAQ-en.md) ?
- [X] Checked the logs for errors in the terminal console ?
- [X] Searched for [similar reports already](https://github.com/universam1/iSpindel/issues) ?

## Report

After updating the firmware of my iSpindel from 6.5.1 to 7.0.0 the d1 mini does not connect to my WIFI properly.

On the initial configuration of the wifi it seems to connect (see logs below).

On the wakups for reporting to ubidots somehow the connection is lost and it's restored. But the data are not sent do ubidots and on the next cycle it's restoring the connection again.

I flashed the old 6.5.1 version again and wifi worked like charm. I also changed SSID & PW of my network -> no fix here.

### Console Logs

On saving WIFI preferences in web ui:
```
[...]
16:39:20.232 -> *WM: Sent wifi save page
16:39:22.230 -> *WM: Connecting to new AP
16:39:22.230 -> *WM: Connecting wifi with new parameters...
16:39:22.269 -> *WM: previous settings invalidated
16:39:22.447 -> *WM: After waiting...
16:39:25.651 -> *WM: 3.21
16:39:25.651 -> *WM: seconds
16:39:25.651 -> *WM: Connection result: 
16:39:25.651 -> *WM: WL_CONNECTED
16:39:25.651 -> saving config...
16:39:25.721 -> {"Name":"iSpindel000","Token":"xxx,"Sleep":15,"Server":"","API":0,"Port":80,"Channel":0,"URI":"","DB":"ispindel","Username":"","Password":"","Job":"ispindel","Instance":"000","Hassio":false,"Vfact":191.8,"TS":0,"OWpin":12,"POLY":"-0.00031*tilt^2+0.557*tilt-14.054","SSID":"ispindodal2","PSK":"xxx","Offset":[0,0,0,0,0,0]}
16:39:25.759 -> saved successfully
16:39:27.969 -> Samples:30 min:90.14 max:91.02 time:2193
16:39:27.969 -> x: -12288 y: 11172 z: -200
[...]
```

When waking up to report to ubidots:
```
[...]
[...]

PR #513, opened 2021-06-15
Added enableWiFiAtBootTime() to enable legacy wifi persistence

peterschrott:
Fixing issue: https://github.com/universam1/iSpindel/issues/507

From version 3 of Arduino core (`ESP8266Wifi`) the persistence of wifi settings has changed. 
Source: https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/generic-class.html#persistent

Adding `enableWiFiAtBootTime()` restores the legacy behavior.

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -1212,6 +1212,12 @@ void setup()
   initDS18B20();
   initAccel();
 
+  #ifdef WIFI_IS_OFF_AT_BOOT
+    // persistence is disabled by default and WiFi does not start automatically at boot
+    // source: https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/generic-class.html#persistent
+    enableWiFiAtBootTime();
+  #endif
+
   // decide whether we want configuration mode or normal mode
   if (shouldStartConfig(validConf))
   {


pio/src/iSpindel.cpp
```diff
@@ -1212,6 +1212,12 @@ void setup()
   initDS18B20();
   initAccel();
 
+  #ifdef WIFI_IS_OFF_AT_BOOT
```
universam1, line 1215: guess we can remove this guard, otherwise it will not be enabled now, right

universam1 (changes requested):
👍🏻

universam1:
good discovery, thank you @peterschrott

pio/src/iSpindel.cpp
peterschrott: Not sure if it can be removed, as the function `WIFI_IS_OFF_AT_BOOT ` is not available in the API of older arduino versions.  `WIFI_IS_OFF_AT_BOOT` is set and introduced with v.3.0 when the persistence behaviour has changed and the function `WIFI_IS_OFF_AT_BOOT ` was added. The example in the docs (link above) show it like that.

universam1 (approved):

tjhowse:
Can confirm - My Wemos D1 mini based iSpindel stopped automatically connecting to wifi after deep sleep after I upgraded it to v7.0.0. I built and flashed v7.1.0 and now it's back to normal. Thanks @peterschrott !

peterschrott:
@tjhowse ur welcome.
But there is not "official" 7.1.0. @universam1 when do you release a new fw?

tjhowse:
I'm just a user, not a maintainer. I don't know when v7.1.0 is going to be promoted from beta.

peterschrott:
Question was going out to the maintainer. But just saw the "pre-release" now. 
Great! I need to brew soon! :)

tjhowse:
Oh sorry! I didn't notice your maintainer @mention. Now I feel silly 😆

universam1:
Didn't have time to test it yet, but based on positive feedback I can promote it to be a release

PR #515, opened 2021-06-23
Extended german parts docu, updated en and zh parts docu

peterschrott:


universam1 (approved):
thank you @peterschrott very much for improving the docs!

PR #528, opened 2021-07-30
Pull my_ variables written to the JSON into a single struct

thegreatgunbantoad:


thegreatgunbantoad:
Just an idea from when I was trying to see if I could iterate through a struct to write to the JSON.  Also this is not Python.  It's still helpful to have a struct as it allows intellisense to show you what's in there.  Adds no functionality whatsoever.

universam1 (approved):
I like it, I started to work on using more structs but did not release. This looks fine to me, refactoring one part a time.

Thanks!

thegreatgunbantoad:
Nice one.  Thank you.

PR #529, opened 2021-07-30
Encapsulation of functions

thegreatgunbantoad:
tempScaleLabel updated to use parameters rather than refer to external variables
scaleTemperature updated to use parameters rather than refer to external variables

Note change of name of scaleTemperature to scaleTemperatureFromC

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -75,7 +75,7 @@ uint32_t DSreqTime = 0;
 int16_t ax, ay, az;
 float Volt, Temperatur, Tilt, Gravity;
 
-float scaleTemperature(float t, uint8_t tempscale)
+float scaleTemperatureFromC(float t, uint8_t tempscale)
 {
   if (tempscale == TEMP_CELSIUS)
     return t;
@@ -569,7 +569,7 @@ bool uploadData(uint8_t service)
   if (service == DTUbiDots)
   {
     sender.add("tilt", Tilt);
-    sender.add("temperature", scaleTemperature(Temperatur, myData.my_tempscale));
+    sender.add("temperature", scaleTemperatureFromC(Temperatur, myData.my_tempscale));
     sender.add("battery", Volt);
     sender.add("gravity", Gravity);
     sender.add("interval", myData.my_sleeptime);
@@ -584,7 +584,7 @@ bool uploadData(uint8_t service)
   {
     sender.add("name", myData.my_name);
     sender.add("tilt", Tilt);
-    sender.add("temperature", scaleTemperature(Temperatur, myData.my_tempscale));
+    sender.add("temperature", scaleTemperatureFromC(Temperatur, myData.my_tempscale));
     sender.add("battery", Volt);
     sender.add("gravity", Gravity);
     sender.add("interval", myData.my_sleeptime);
@@ -599,7 +599,7 @@ bool uploadData(uint8_t service)
   if (service == DTMQTT)
   {
     sender.add("tilt", Tilt);
-    sender.add("temperature", scaleTemperature(Temperatur, myData.my_tempscale));
+    sender.add("temperature", scaleTemperatureFromC(Temperatur, myData.my_tempscale));
     sender.add("temp_units", tempScaleLabel(myData.my_tempscale));
     sender.add("battery", Volt);
     sender.add("gravity", Gravity);
@@ -614,7 +614,7 @@ bool uploadData(uint8_t service)
   if (service == DTTHINGSPEAK)
   {
     sender.add("tilt", Tilt);
-    sender.add("temperature", scaleTemperature(Temperatur, myData.my_tempscale));
+    sender.add("temperature", scaleTemperatureFromC(Temperatur, myData.my_tempscale));
     sender.add("temp_units", tempScaleLabel(myData.my_tempscale));
     sender.add("battery", Volt);
     sender.add("gravity", Gravity);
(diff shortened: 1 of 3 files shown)


thegreatgunbantoad:
Again not a functional update.  Not sure why linting is failing though.

pppedrillo:
Why?

thegreatgunbantoad:
> Why?

It makes code easier to reuse and understand.  Would also allow you to to pull the functions out of the main file and put them in a library.  A task it a bit different, like an initialisation task, but functions should generally be properly parameterised.

pppedrillo:
> It makes code easier to reuse and understand. Would also allow you to to pull the functions out of the main file and put them in a library. A task it a bit different, like an initialisation task, but functions should generally be properly parameterised.

"Parameterization" != "to duplicate all the used global/static variables using arguments"

Why only those two functions then? Follow your logic, you'll need to refactor most of the rest of the functions which also use global variables (voltage, tilt etc etc) in the same manner, no?
And what if function use 125 global vars? Should it have 125 arguments?

And generally copying arguments on every function call causes code run slower+ extra mem use on stack - you are passing by value , not using a references.

pio/src/iSpindel.cpp
```diff
@@ -75,25 +75,25 @@ uint32_t DSreqTime = 0;
 int16_t ax, ay, az;
 float Volt, Temperatur, Tilt, Gravity;
 
-float scaleTemperature(float t)
+float scaleTemperatureFromC(float t, uint8_t tempscale)
 {
-  if (myData.my_tempscale == TEMP_CELSIUS)
+  if (tempscale == TEMP_CELSIUS)
```
pppedrillo, line 80: Why do not use switch here?

pio/src/iSpindel.cpp
```diff
@@ -75,25 +75,25 @@ uint32_t DSreqTime = 0;
...
     return t + 273.15f;
   else
     return t; // Invalid value for myData.my_tempscale => default to celsius
 }
 
-String tempScaleLabel(void)
+String tempScaleLabel(uint8_t tempscale)
 {
-  if (myData.my_tempscale == TEMP_CELSIUS)
+  if (tempscale == TEMP_CELSIUS)
```
pppedrillo, line 92: Same as above - why not switch?

pppedrillo (changes requested):

pio/src/iSpindel.cpp
thegreatgunbantoad: Could do, I wasn't intending to make great swathes of changes all at once.

thegreatgunbantoad:
> > It makes code easier to reuse and understand. Would also allow you to to pull the functions out of the main file and put them in a library. A task it a bit different, like an initialisation task, but functions should generally be properly parameterised.
> 
> "Parameterization" != "to duplicate all the used global/static variables using arguments"
> And those functions still not in the library or taken out of main file.
> 
> Generally why only those two functions were picked? Follow your logic, you'll need to refactor most of the rest of the functions which also use global variables (voltage, tilt etc etc) in the same manner, no?
> And what if function use 125 global vars? Should it have 125 arguments?
> 
> I'd also advise to pass args by reference whenever it makes sense (like here, myData.my_tempscale) and master a switch statement (instead of the if-else-if-else... spaghetti as soon as it is refactoring)
> 
> Renaming a function to "FromC" looks good.

Only picked those two as it was just a start at getting the functions parametrised.  If a function uses lots of global variables I think it might need re-writing.
Note the function that writes to the JSON could now just be passed the myData struct.  I was really hoping C++ had an iterator like Python that would have made writing to and reading from the JSON slicker.

Yes the plan was to refactor as many functions as is practical.  Globals are generally best avoided if reasonable, they can make code very hard to follow.

I sort of prefer pass by value as you know the variable can't change while the function is operating and the function can't alter the inputs.  But if you wanted several outputs then pass by ref is handy.  Is there a processor advantage to ref over val?

Agree with using a switch.

pppedrillo:
> I was really hoping C++ had an iterator like Python that would have made writing to and reading from the JSON slicker.
> 

There is nothing to do with C++ language, but rather the concrete implementation of JSON (de)/serializer which is not a part of C++ standards.

> Yes the plan was to refactor as many functions as is practical. 

Thats my main concern. 
It can be a pull request then all the changes are done, not while work in progress. 

>Globals are generally best avoided if reasonable, they can make code very hard to follow.

There are some more serious problems with globals and statics, not the only code readability.

> I sort of prefer pass by value as you know the variable can't change while the function is operating and the function can't alter the inputs. But if you wanted several outputs then pass by ref is handy. Is there a processor advantage to ref over val?
> 

Passing by val not always guarantees what some data can't be changed.
If you pass an object by value and this object has a pointer to data, you still can change a data by this pointer.

Omitting an extra copy is one of advantages. Not a huge deal with one shot functions and PODs, but may cause performance issues if function gets called frequently and you pass a huge object(s) to it.

PjotrekSE:
I make a struct of my globals, and pass the struct by reference if I need
to change a global, and by value if not.

universam1 (approved):
Thanks 👍🏻

universam1:
@thegreatgunbantoad there are merge conflicts, mind rebasing please?

thegreatgunbantoad:
@universam1 I'll have look later, not great at the git thing yet, so might need a hand.

pppedrillo:
> 
> 
> @universam1 I'll have look later, not great at the git thing yet, so might need a hand.

I rebased those changes, #543 @universam1

thegreatgunbantoad:
@pppedrillo does that mean I can just close this request then?  Sorry full of a cough and such so my brain is struggling.

pppedrillo:
> 
> 
> @pppedrillo does that mean I can just close this request then? Sorry full of a cough and such so my brain is struggling.

@thegreatgunbantoad  Yes, if you are ok with changes from #543

thegreatgunbantoad:
@pppedrillo frazzled brain thinks looks good.

PR #531, opened 2021-08-12
make tilt angle independent of the board orientation

m0n5t3r:
this solves #445 (yes, the calibration will account for the weird numbers, but it's nice to have a number one can relate to - for instance, instructions say the 1.000 angle should be 25 degrees - with the board inverted that would be 155 degrees, having to remember to subtract from 180 degrees is less than ideal :) )

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -945,7 +945,7 @@ float calculateTilt()
   if (ax == 0 && ay == 0 && az == 0)
     return 0.f;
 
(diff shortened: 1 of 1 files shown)


ErikdBr:
Wasn't this already resolved in 6.5.1 ?

m0n5t3r:
apparently not; I updated my ispindels a few months ago and only got to use one  recently; initially I thought I calibrated it wrong or something, but then I did the trig on paper and it checked out (negative z will result in the supplementary angle, i.e. 180 - our tilt angle)

universam1 (approved):
👍🏻

universam1:
Thanks for that

PR #541, opened 2021-09-29
Some useful docs

thegreatgunbantoad:
Cherry Philip V4 Build Guide
Rapid Start Guide
Quick Start Guide


universam1 (changes requested):
Thank you for the contributions!

In general I steer away from using binary document formats other than html files for various reasons. They are not commonly accepted in modern web deployments as they turn to be stale or "dead" as they are hard to keep updated, limited support for mobile devices and people restrain to open them because of virus fears. Also it is impossible to review them here.
I believe it shouldn't be too hard to convert them to html files instead which would be fine to merge then.

Thanks!

PR #559, opened 2021-12-13
Support SG on BierBot Bricks

BernhardSchlegel:
If iSpindel transmits SG (and not °P) 1 decimal (1.0, `round(x*10)/10`) is not sufficient. This PR ensures, that 4 decimals are send to the BierBot Bricks backend. This way, gravity values like SG=1.0482 can be transmitted to the BierBot Bricks backend without loss of information.

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -788,7 +788,7 @@ bool uploadData(uint8_t service)
     sender.add("brand", "wemos_d1_mini");
     sender.add("version", FIRMWAREVERSION);
     sender.add("chipid", chipidHashed);
-    sender.add("s_number_wort_0", (float)(round(Gravity * 1000) / 1000));
+    sender.add("s_number_wort_0", (float)(round(Gravity * 10000) / 10000));
     sender.add("s_number_temp_0", (float)(round(Temperatur * 10) / 10)); // always transmit °C
     sender.add("s_number_voltage_0", (float)round(Volt * 100) / 100);
     sender.add("s_number_wifi_0", WiFi.RSSI());
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -788,7 +788,7 @@ bool uploadData(uint8_t service)
     sender.add("brand", "wemos_d1_mini");
     sender.add("version", FIRMWAREVERSION);
     sender.add("chipid", chipidHashed);
-    sender.add("s_number_wort_0", (float)(round(Gravity * 10) / 10));
+    sender.add("s_number_wort_0", (float)(round(Gravity * 1000) / 1000));
     sender.add("s_number_temp_0", (float)(round(Temperatur * 10) / 10)); // always transmit °C
     sender.add("s_number_voltage_0", (float)round(Volt * 100) / 100);
     sender.add("s_number_wifi_0", WiFi.RSSI());
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -788,11 +788,11 @@ bool uploadData(uint8_t service)
     sender.add("brand", "wemos_d1_mini");
     sender.add("version", FIRMWAREVERSION);
     sender.add("chipid", chipidHashed);
-    sender.add("s_number_wort_0", (float)(round(Gravity * 10000) / 10000));
-    sender.add("s_number_temp_0", (float)(round(Temperatur * 10) / 10)); // always transmit °C
-    sender.add("s_number_voltage_0", (float)round(Volt * 100) / 100);
+    sender.add("s_number_wort_0", Gravity); // gravity can be in SG or °P, depending on user setting
+    sender.add("s_number_temp_0", Temperatur); // always transmit °C
+    sender.add("s_number_voltage_0", Volt);
     sender.add("s_number_wifi_0", WiFi.RSSI());
-    sender.add("s_number_tilt_0", (float)(round(Tilt * 100) / 100));
+    sender.add("s_number_tilt_0", Tilt);
 
     CONSOLELN(F("\ncalling BRICKS"));
 
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -788,7 +788,7 @@ bool uploadData(uint8_t service)
     sender.add("brand", "wemos_d1_mini");
     sender.add("version", FIRMWAREVERSION);
     sender.add("chipid", chipidHashed);
-    sender.add("s_number_wort_0", Gravity); // gravity can be in SG or °P, depending on user setting
+    sender.add("s_number_wort_0", Gravity);    // gravity can be in SG or °P, depending on user setting
     sender.add("s_number_temp_0", Temperatur); // always transmit °C
     sender.add("s_number_voltage_0", Volt);
     sender.add("s_number_wifi_0", WiFi.RSSI());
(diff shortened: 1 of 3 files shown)


jonathanschneider (dismissed):
Commit changes decimal point of gravity measurement sent to BierBot bricks to support SG and has no effect on the rest of the firmware.

pppedrillo:
4 decimals in SG can be a bit of overshoot.

BernhardSchlegel:
That's one more than actually needed to be able to round backend-wise and enable data freshness checks. I'm happy to change it if really necessary. But IMHO it should not hinder the merge.

WireTownMan (approved):
This change corrects a defect that allows bierbots to use Plato or SG.  Does not impact anything other than number of decimal places transmitted

thegreatgunbantoad:
API_BRICKS seems to be the only API that does this.  It this truncation needed at all?

BernhardSchlegel:
I'm personally a fan of a deterministic memory footprint in non-garbage collected environments and deterministic POST payloads. But: I don't want to force other APIs to follow this approach. 

Is a discussion about one decimal more or less really worth our time? I would really appreciate if we come to an agreement :) If that's the only concern hindering the merge, I'm happy to change it to avoid further waste of time and energy on all sides.

pppedrillo:
> Is a discussion about one decimal more or less really worth our time?  I would really appreciate if we come to an agreement

This discussion at least can help to save this pull request from the stale bot's shredder :) Apparently, you haven't looked at these things from this point of view :)

BernhardSchlegel:
hehe, that's indeed correct. So how do we proceed? Can someone with the necessary rights to merge a PR make a clear statement ("remove truncation"/ "truncation to three decimals is fine" / "trundcation to four decimals is fine")?

Thanks in advance!

WireTownMan:
> > Is a discussion about one decimal more or less really worth our time?  I would really appreciate if we come to an agreement
> 
> This discussion at least can help to save this pull request from the stale bot's shredder :) Apparently, you haven't looked at these things from this point of view :)

It is all about decimal places.  With 1 decimal place bierbots will only support Plato units,  with 3 or more decimal places it will support Plato and Specific Gravity (SG).  So, yes this discussion is all about one the number of decimal places transmitted to the bierbots API

pppedrillo:
> It is all about decimal places. With 1 decimal place bierbots will only support Plato units, with 3 or more decimal places it will support Plato and Specific Gravity (SG). So, yes this discussion is all about one the number of decimal places transmitted to the bierbots API

To me it sounds like an attempt to fix bierbots issue (eg unable to configure gravity units in bierbots) on the iSpindel side?

WireTownMan:
> > It is all about decimal places. With 1 decimal place bierbots will only support Plato units, with 3 or more decimal places it will support Plato and Specific Gravity (SG). So, yes this discussion is all about one the number of decimal places transmitted to the bierbots API
> 
> To me it sounds like an attempt to fix bierbots issue (eg unable to configure gravity units in bierbots) on the iSpindel side? Why dont fix it on bierbots side?

Fixing on the Bierbots side would force ispindel to be a device that is setup for Plato units.  This is inconsistent with other interfaces that allow the gravity to be either P or SG.

BernhardSchlegel:
> Overall, to me it sounds like an attempt to fix bierbots issue (eg unable to configure gravity units in bierbots) on the iSpindel side? Why dont fix it on bierbots side?

Nope, BierBot Bricks backend can't do much if the iSpindle user decided to go with SG and receives a `1.0` reading...

But again: I'm happy to fullfill any request to be able to merge. Please just tell me what you want.

pppedrillo:
@BernhardSchlegel 
Sorry I feel I still missing something.
Putting aside deterministic things, what if there would be no truncation at all? 
I mean from the very beginning  - in your very first bierbots support version. If all the data which currently truncated (SG, Temp, Voltage and tilt) would be sent as is (eg their "raw" values), as all the other APIs do? 
If truncation (or any other conversion) need, it must be done on the bierbots side?
And no need to put a magic numbers 10, 10000 etc and change them to "fix" something?

thegreatgunbantoad:
I think it may be worth splitting this into two items.  One quick and easy fix as this current request has.  And then maybe open a discussion point as to whether or not the truncation can be removed altogether.
From what I can see the current correction looks fine, and I don't see a reason not to merge it.
Long term I do think we need to answer whether or not the raw value could be used.

thegreatgunbantoad (approved):
This looks fine, but please open a discussion point to raise the issue of whether or not BierBot Bricks could simply accept the raw untruncated value.

pppedrillo:
> From what I can see the current correction looks fine, and I don't see a reason not to merge it.

This merge can potentially cause future "fixes" in the same style. 
The problem was with initial implementation which has such truncation from the beginning.

If there is no technical showstoppers for sending raw values without truncation, I'd propose change values to raw now and forget about bierbots fixing for a while.  @thegreatgunbantoad

pio/src/iSpindel.cpp
```diff
@@ -788,7 +788,7 @@ bool uploadData(uint8_t service)
     sender.add("brand", "wemos_d1_mini");
     sender.add("version", FIRMWAREVERSION);
     sender.add("chipid", chipidHashed);
-    sender.add("s_number_wort_0", (float)(round(Gravity * 10) / 10));
+    sender.add("s_number_wort_0", (float)(round(Gravity * 10000) / 10000));
```
pppedrillo, line 791: Can we consider to send a raw values (without truncation) instead of such fixes with changing a magic numbers?

pppedrillo (changes requested):
I'd propose to get rid of truncation and send a raw values in the same way as the other APIs do.

pio/src/iSpindel.cpp
BernhardSchlegel: done! 👍

pppedrillo (approved):
LGTM
Clang format has to be run onto iSpindel.cpp to get rid of lint errors.

pppedrillo (approved):

pppedrillo (approved):
Can be merged as soon as build gets fixed,

thegreatgunbantoad:
> > From what I can see the current correction looks fine, and I don't see a reason not to merge it.
> 
> This merge can potentially cause future "fixes" in the same style. The problem was with initial implementation which had such truncation from the beginning.
> 
> If there is no technical showstoppers for sending raw values without truncation, I'd propose change values to raw now and forget about bierbots fixing for a while. @thegreatgunbantoad

The sort of point I was going for was that this was a fix in the style of the original code which I can't see any major issue with.

Secondary to that was the query of weather the original should have been coded that way which is a bit of a different issue and could have been made it's own pull request in order to keep the two issues segregated.

pppedrillo:
> Secondary to that was the query of weather the original should have been coded that way which is a bit of a different issue and could have been made it's own pull request in order to keep the two issues segregated.

Martin, this is not a "two issues" (which should be kept segrregated) but the only one and the same. Nevertheless, @BernhardSchlegel already removed truncation. @thegreatgunbantoad

thegreatgunbantoad:
> > Secondary to that was the query of weather the original should have been coded that way which is a bit of a different issue and could have been made it's own pull request in order to keep the two issues segregated.
> 
> Martin, looks like this is not "two issues" (which should be kept segrregated) but the only one and the same. Nevertheless, @BernhardSchlegel already removed truncation. @thegreatgunbantoad

Indeed.  I still don't think they needed to be the same problem (solve one thing at at time or you'll never get anywhere), but one did certainly raise the other.  But that looks moot now anyway.  Cheers to @BernhardSchlegel for losing his mind while we all discuss the awkwardness.

thegreatgunbantoad:
@pppedrillo huh, didn't realise git hub knew my name.

thegreatgunbantoad:
@BernhardSchlegel, oh cool are you the BierBot-Bricks developer, didn't realise that.  That's cool.

BernhardSchlegel:
Can we please merge this?

BernhardSchlegel:
Thanks!

PR #572, opened 2022-03-02
Ability to save and load data of calibration points.

vitotai:
I've wanted to add this functionality long time ago.


universam1 (approved):
Thank you for the contribution

PR #579, opened 2022-03-31
docs: correct antenna length in docs/FAQ-en.md

PharmaJap:
The forum post linked to in the FAQ specifies an antenna length of 32mm, not 3.2mm. I verified the math, and lambda/4 for the 2.4GHz band is indeed ~32mm.


universam1 (approved):
👍🏻

PR #580, opened 2022-04-14
Adding support for InfluxDB in HTTPS mode

PharmaJap:
For security purposes, it is inadvisable to run InfluxDB on the public-facing internet in HTTP mode (for example, on a VPS).

This commit adds a checkbox to the InfluxDB service to enable HTTPS, for use with servers that are configured for it.

Compiled and running/sending successfully in both HTTP and HTTPS mode on a Lolin Wemos D1 Mini v4.

### DIFF
--- a/pio/lib/Sender/Sender.cpp
+++ b/pio/lib/Sender/Sender.cpp
@@ -459,7 +459,7 @@ bool SenderClass::sendGenericPost(String server, String uri, uint16_t port)
   return true;
 }
 
-bool SenderClass::sendInfluxDB(String server, uint16_t port, String db, String name, String username, String password)
+bool SenderClass::sendInfluxDB(String server, uint16_t port, String db, String name, String username, String password, bool usehttps)
 {
   HTTPClient http;
 
@@ -468,7 +468,16 @@ bool SenderClass::sendInfluxDB(String server, uint16_t port, String db, String n
 
   CONSOLELN(String(F("INFLUXDB: posting to db: ")) + uri);
   // configure traged server and uri
-  http.begin(_client, server, port, uri);
+
+  if (usehttps == true)
+  {
+    _secureClient.setInsecure();
+    http.begin(_secureClient, server, port, uri);
+  }
+  else
+  {
+    http.begin(_client, server, port, uri);
+  }
 
   if (username.length() > 0)
   {
--- a/pio/lib/Sender/Sender.h
+++ b/pio/lib/Sender/Sender.h
@@ -25,7 +25,7 @@ public:
   bool sendThingSpeak(String token, long Channel);
   bool sendGenericPost(String server, String uri, uint16_t port = 80);
   bool sendHTTPSPost(String server, String uri);
-  bool sendInfluxDB(String server, uint16_t port, String db, String name, String username, String password);
+  bool sendInfluxDB(String server, uint16_t port, String db, String name, String username, String password, bool usehttps);
   bool sendPrometheus(String server, uint16_t port, String job, String instance);
   bool sendUbidots(String token, String name);
   bool sendMQTT(String server, uint16_t port, String username, String password, String name);
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -69,6 +69,7 @@ iData myData;
 //myData.hassio = false;
 bool hassio_changed = false;
 #endif
+bool usehttps_changed = false;
 
 uint32_t DSreqTime = 0;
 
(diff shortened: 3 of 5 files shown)


pio/src/iSpindel.cpp
```diff
@@ -646,6 +646,22 @@ bool uploadData(uint8_t service)
 
     return sender.sendInfluxDB(myData.server, myData.port, myData.db, myData.name, myData.username, myData.password);
   }
+  if (service == DTInfluxDBHTTPS)
+  {
+    sender.add("tilt", Tilt);
```
pppedrillo, line 651: All the copy-paste started from this line can be omitted. `code`
`
#if API_INFLUXDB
  if (service == DTInfluxDB **|| service == DTInfluxDBHTTPS**)
  {
    sender.add("tilt", Tilt);
    sender.add("temperature", scaleTemperatureFromC(Temperatur, myData.tempscale));
    sender.add("temp_units", tempScaleLabel(myData.tempscale));
    sender.add("battery", Volt);
    sender.add("gravity", Gravity);
    sender.add("interval", myData.sleeptime);
    sender.add("RSSI", WiFi.RSSI());
    CONSOLELN(F("\ncalling InfluxDB"));
    CONSOLELN(String(F("Sending to db: ")) + myData.db + String(F(" w/ credentials: ")) + myData.username +
              String(F(":")) + myData.password);
   
**if (service == DTInfluxDB)
{
   return sender.sendInfluxDB(myData.server, myData.port, myData.db, myData.name, myData.username, myData.password);
}
else
{
    return sender.sendInfluxDBHTTPS(myData.server, myData.port, myData.db, myData.name, myData.username,
                                    myData.password);
}**
`

pio/lib/Sender/Sender.cpp
```diff
@@ -525,6 +525,75 @@ bool SenderClass::sendInfluxDB(String server, uint16_t port, String db, String n
   return true;
 }
 
+bool SenderClass::sendInfluxDBHTTPS(String server, uint16_t port, String db, String name, String username, String password)
+{
+  std::unique_ptr<BearSSL::WiFiClientSecure> client(new BearSSL::WiFiClientSecure);
+  client->setInsecure();
+
```
pppedrillo, line 532: Looks like SenderClass::sendInfluxDBHTTPS differs from SenderClass::sendInfluxDB by only lines 530 and 531 (and line 540 where you refer to client). 
Was it worth to copypaste the rest of the SenderClass::sendInfluxDB implementation?
May be simply to introduce a boolean flag bUseHTTPS to SenderClass::sendInfluxDB arguments and either create new local client or use existing global one ?

pio/lib/Sender/Sender.h
```diff
@@ -26,6 +26,7 @@ class SenderClass
   bool sendGenericPost(String server, String uri, uint16_t port = 80);
   bool sendHTTPSPost(String server, String uri);
   bool sendInfluxDB(String server, uint16_t port, String db, String name, String username, String password);
+  bool sendInfluxDBHTTPS(String server, uint16_t port, String db, String name, String username, String password);
```
pppedrillo, line 29: bool sendInfluxDB(String server, uint16_t port, String db, String name, String username, String password, **bool bUseHTTPS**); ?

pppedrillo (changes requested):
Perhaps, there is a better way to keep code clean and get rid of dreaded copy-paste spaghetti.

pio/src/iSpindel.cpp
PharmaJap: Reimplemented to use a bool/checkbox, as suggested.

pio/lib/Sender/Sender.cpp
PharmaJap: Reimplemented as a bool as suggested. There is still some slight duplication due to differing function names, but I tried to keep it to a minimum.

pio/lib/Sender/Sender.h
PharmaJap: Reimplemented to use a bool/checkbox, as suggested.

pppedrillo (changes requested):
Looks good. But imo can be even better. Pls see my proposals for getting rid of the last copypaste block.

PharmaJap:
Updated with an obvious solution that I missed; no spaghetti. Might be ready to ship?

pppedrillo:
@PharmaJap Lets tag @universam1

universam1 (approved):
Thanks, looks clean and good to maintain

PR #581, opened 2022-05-05
Added support for MPU6500 basic functionality

Bananamannn:
Will need to be double checked using a MPU6050 as I don't have one on hand. However MPU6500 seems to work just fine

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -971,13 +971,23 @@ float calculateTilt()
 bool testAccel()
 {
   uint8_t res = Wire.status();
-  if (res != I2C_OK)
+    if (res != I2C_OK)
     CONSOLELN(String(F("I2C ERROR: ")) + res);
-
-  bool con = accelgyro.testConnection();
+  
+  bool con = false;
+  if (accelgyro.getDeviceID()==0x34)
+  {
+    con = true;
+  }
+  
+  if (accelgyro.getDeviceID()==0x38)
+  {
+    con = true;
+  }
+  
   if (!con)
     CONSOLELN(F("Acc Test Connection ERROR!"));
-
+      
   return res == I2C_OK && con == true;
 }
 


pio/src/iSpindel.cpp
```diff
@@ -971,13 +971,23 @@ float calculateTilt()
...
+  bool con = false;
+  if (accelgyro.getDeviceID()==0x34)
+  {
+    con = true;
+  }
+  
+  if (accelgyro.getDeviceID()==0x38)
+  {
+    con = true;
+  }
```
universam1, line 986:
```suggestion
  auto id = accelgyro.getDeviceID();
  if ( id ==0x34|| id ==0x38 )
    con = true;
```

universam1: Please add a comment to which ID what device type matches

universam1 (changes requested):

pio/src/iSpindel.cpp
Bananamannn: Updated. I like your update to the code

mp-se:
You need to patch the MPU6050.cpp file as well or it will fail on this function; accelgyro.testConnection()

Bananamannn:
> You need to patch the MPU6050.cpp file as well or it will fail on this function; accelgyro.testConnection(). This code will work;
> 
> ` bool MPU6050_Base::testConnection() { return getDeviceID() == 0x34 || getDeviceID() == 0x38; // Allow both MPU6050 and MPU6000 }`

I didn't want to make any changes to the standard MPU6050 repository so I tried to remove the "accelgyro.testConnection()" line in the main ISpindel.cpp file and replace it so that function was never used. Did I miss it somewhere else other than the repo?

mp-se:
> > You need to patch the MPU6050.cpp file as well or it will fail on this function; accelgyro.testConnection(). This code will work;
> > ` bool MPU6050_Base::testConnection() { return getDeviceID() == 0x34 || getDeviceID() == 0x38; // Allow both MPU6050 and MPU6000 }`
> 
> I didn't want to make any changes to the standard MPU6050 repository so I tried to remove the "accelgyro.testConnection()" line in the main ISpindel.cpp file and replace it so that function was never used. Did I miss it somewhere else other than the repo?

Sorry, my misstake. didnt look at the changes. just noticed that there was a call to that method in the code.

universam1 (changes requested):
Linter : code formatter fails

thegreatgunbantoad:
Is it worth altering that if statement a bit to as to not rely on operation order i.e.:
 ` if ( (id == 0x34) || (id == 0x38) )`

universam1:
> Is it worth altering that if statement a bit by adding brackets so as to not rely on operation order i.e.: ` if ( (id == 0x34) || (id == 0x38) )`

https://docs.microsoft.com/en-us/cpp/c-language/precedence-and-order-of-evaluation?view=msvc-170 parentheses are not necessary here

thegreatgunbantoad:
> 

I know they aren't strictly needed, it's more of a good practise / style kind of thing, also for readability.

Bananamannn:
> Linter : code formatter fails

I honestly have no idea what this is. I haven't changed this from your master branch. Any ideas why it's failing on here but not when I build it on my PC?

Bananamannn (commented):
Changes made

Bananamannn:
Could I please get some advice on how to deal with the Linter fail, any pointer would be greatly appreciated. I honestly have no idea why it keeps failing.

https://github.com/universam1/iSpindel/runs/6349107436?check_suite_focus=true

thegreatgunbantoad:
Is the dark green stuff in the files changed bit the stuff the linter doesn't like?  I can't find the results of the linter.
So possibly a space after the 2nd comma in main.yml line 3
and one less space before the if in ispindel.cpp line 974

Bananamannn:
It's in the link below. I added the linter to my fork for testing.
The red and green are the changes that i've made.

[https://github.com/Bananamannn/iSpindel/runs/6418152846?check_suite_focus=true](https://github.com/Bananamannn/iSpindel/runs/6418152846?check_suite_focus=true)

image

pppedrillo:
@Bananamannn All you had to to to get rid of format errors is just execute "Format document" command in VS Code

Capture

Bananamannn:
> @Bananamannn All you had to to to get rid of format errors is just execute "Format document" command in VS Code (assuming you have clang-format installed by default)
> 
> Capture

Oh right, thanks for the info. I'll defiantly do it that way next time. I did it all through the Github editor this round.

Bananamannn (commented):
Fixed formatting. Linter now passes

universam1 (approved):
+1
Excellent @Bananamannn thanks for contribution

PR #591, opened 2022-08-19
Add timeout when waiting for gyro data ready status

stuartraetaylor:
Adds a timeout to the loop that waits for `accelgyro.getIntDataReadyStatus()`.

After 30 iterations the loop (same as wifi check), breaks and logs. Whatever readings we get from the gyro are logged as usual.

I recently had a faulty MPU-6050 that hung in this loop -
```
    while (!accelgyro.getIntDataReadyStatus())
       delay(2);
```

The iSpindel in this state doesn't send out any readings or log anything to serial. The idea of this PR is to make an MPU-6050 failure easier to troubleshoot in this situation.

### DIFF
--- a/pio/src/iSpindel.cpp
+++ b/pio/src/iSpindel.cpp
@@ -997,8 +997,18 @@ float getTilt()
 
   for (; i < MEDIANROUNDSMAX; i++)
   {
+    uint8_t wait = 0;
     while (!accelgyro.getIntDataReadyStatus())
+    {
       delay(2);
+
+      if (++wait > 50)
+      {
+        CONSOLELN(F("Error: timed out waiting for gyro!"));
+        break;
+      }
+    }
+
     getAccSample();
     float _tilt = calculateTilt();
     samples.add(_tilt);


pio/src/iSpindel.cpp
```diff
@@ -997,8 +997,18 @@ float getTilt()
 
   for (; i < MEDIANROUNDSMAX; i++)
   {
+    uint8_t wait = 0;
     while (!accelgyro.getIntDataReadyStatus())
+    {
       delay(2);
+
+      if (++wait > 50)
+          break;
```
universam1, line 1006: The error print can be here before the break and hence the second conditional can be omitted then

universam1 (changes requested):
I like it! One small change and it is good to go. Thanks

pio/src/iSpindel.cpp
stuartraetaylor: Yes! Updated, thanks.

universam1 (approved):
👍

universam1:
Thank you

PR #598, opened 2023-01-02
Repair for ArduinoJson 6.20.0

piv3rt:
I couldn't compile the project and experienced the following error:
`pio/lib/Sender/Sender.cpp:499:31: error: no matching function for call to 'ArduinoJson6200_F1::JsonVariant::is<char*>()'`

I browsed ArduinoJson's documentation and came upon the following line on [the page documenting the JsonVariant::is<T>()](https://arduinojson.org/v6/api/jsonvariant/is/) function:
`bool is<char*>() const;  // ⚠️ deprecated since 6.18`

After applying the change you will find in this PR, I was able to compile the project and upload it to a NodeMCU but I couldn't fully test the iSpindel as I am missing some sensors at the moment.


universam1 (approved):
+1 Thank you for the fix!

